PR #3, opened 2018-08-16
Fix copyright on pom

jonahgraham:
I am using this primarily to test PR builder on Jenkins -- updating the copyright on the pom is just a side effect :-)


jonahgraham:
ok to test

PR #16, opened 2018-11-10
Fix bug 534270 (Unexpected padding to right of RoundedToolItems when RoundedToolbar has SWT.HIDE_SELECTIO)

lcaron:
When using SWT.RADIO buttons and SWT.HIDE_SELECTION on the toolbar,
there's some unexpected space/padding to the right hand side of the
icon. (perhaps also with other styles, didn't try!)

You can actually see this on the Wiki screenshots at
https://wiki.eclipse.org/Nebula_RoundedToolbar.

We'd ideally want SWT.HIDE_SELECTION to look like the regular TOGGLE
case.


eclipsewebmaster:
Issue tracker reference:
https://bugs.eclipse.org/bugs/show_bug.cgi?id=534270

jonahgraham:
Ok to test

PR #19, opened 2018-11-13
Bug 538599 - property description not wrapped and not visible if long

lcaron:
Fix a problem when description is too long.

Thanks Andrei Diaconu for its bugfix


lcaron:
Canceled

crisnct:
Thank you Laurent for the fix. At the ticket you specified the fix will be available in the version 2.10 of Nebula, but when aproximatley that will be available to download?
Now Nebula is at version 2.01 and it seems we needs to wait a lot until 2.10 will be released

PR #20, opened 2018-11-13
Bug 538599 - property description not wrapped and not visible if long

lcaron:
Fix a problem when description is too long.

Thanks Andrei Diaconu for its bugfix


lcaron:
Canceled (bad author)

lcaron:
Canceled (bad author)

PR #28, opened 2018-11-21
Bug 540967 - Inverted axis tick labelling bug

PeterC-DLS:
When an axis is inverted, the end points are miscalculated leading to strange appearance of tick labels

Modify check for end condition, simplify position calculation, add inverted check to new range, and add unit tests


eclipsewebmaster:
Issue tracker reference:
https://bugs.eclipse.org/bugs/show_bug.cgi?id=540967

wimjongman:
Ok to test

PR #33, opened 2018-12-05
Bug 507140 - SWTException below GalleryItem.setImage

lcaron:
Method setImage:
The check (Widget is Disposed) was done in the super() method.
I propose to check before the call to the super method.


eclipsewebmaster:
Issue tracker reference:
https://bugs.eclipse.org/bugs/show_bug.cgi?id=507140

PR #35, opened 2018-12-16
Bug 542812 : Badged Label Control for Nebula

lcaron:
Initial release


widgets/badgedlabel/org.eclipse.nebula.widgets.badgedlabel/src/org/eclipse/nebula/widgets/badgedlabel/BADGE_COLOR.java
```diff
@@ -0,0 +1,43 @@
...
+package org.eclipse.nebula.widgets.badgedlabel;
+
+import org.eclipse.swt.SWT;
+import org.eclipse.swt.graphics.RGB;
+
+/**
+ * This enumeration provides a set of color for the BadgedLabel control
+ */
+public enum BADGE_COLOR {
+	BLUE(0, 123, 255, SWT.COLOR_WHITE), //
```
mickaelistria: This should probably use the default SWT color symbols or the RGB class instead of introducing a new API.

widgets/badgedlabel/org.eclipse.nebula.widgets.badgedlabel/src/org/eclipse/nebula/widgets/badgedlabel/BADGE_COLOR.java
```diff
@@ -0,0 +1,43 @@
...
+	GREEN(40, 167, 69, SWT.COLOR_WHITE), //
+	RED(220, 53, 69, SWT.COLOR_WHITE), //
+	YELLOW(255, 193, 7, SWT.COLOR_BLACK), //
+	CYAN(23, 162, 184, SWT.COLOR_WHITE), //
+	BLACK(52, 58, 64, SWT.COLOR_WHITE);
+
+	private RGB rgb;
+	private int textColor;
+
+	private BADGE_COLOR(int red, int green, int blue, int textColor) {
```
mickaelistria: One may want to use their own colors, we shouldn't constrain a limited set.

widgets/badgedlabel/org.eclipse.nebula.widgets.badgedlabel/src/org/eclipse/nebula/widgets/badgedlabel/BADGE_COLOR.java
```diff
@@ -0,0 +1,43 @@
...
+		rgb = new RGB(red, green, blue);
+		this.textColor = textColor;
```
mickaelistria: Would it be possible to deduce the textcolor (black/white) dynamically from the background color? That would be one less setting to worry the user about.

widgets/badgedlabel/org.eclipse.nebula.widgets.badgedlabel/src/org/eclipse/nebula/widgets/badgedlabel/BADGE_LOCATION.java
```diff
@@ -0,0 +1,17 @@
...
+ * Contributors: Laurent CARON (laurent.caron at gmail dot com) - initial API and
+ * implementation
+ *******************************************************************************/
+package org.eclipse.nebula.widgets.badgedlabel;
+
+/**
+ * This enumeration describes the possible locations of the badge
+ */
+public enum BADGE_LOCATION {
+	TOP_LEFT, TOP_RIGHT, BOTTOM_LEFT, BOTTOM_RIGHT
```
mickaelistria: Please avoid a dedicated type here and use a combination of SWT.TOP , SWT.LEFT and so on like ControlDecoration is using.

mickaelistria (changes requested):
We should avoid introducing extra types when SWT already has the necessary APIs.
Also, the overlap with ControlDecoration is huge, but the differences are there too (ControlDecoration only support image, and items are hoverable but not clickable...).
Would just adding a `setBackground` and a `setText` to ControlDecoration cover this story in a more unified and centralized way?

widgets/badgedlabel/org.eclipse.nebula.widgets.badgedlabel/src/org/eclipse/nebula/widgets/badgedlabel/BADGE_COLOR.java
lcaron: Enumeration removed and introduce a new setPredefinedColor(int color) that uses predefined SWT colors (SWT.COLOR_BLUE, SWT.COLOR_GRAY...)

widgets/badgedlabel/org.eclipse.nebula.widgets.badgedlabel/src/org/eclipse/nebula/widgets/badgedlabel/BADGE_COLOR.java
lcaron: You can change the colors by using setBadgeBackground and setBadgeForeground

widgets/badgedlabel/org.eclipse.nebula.widgets.badgedlabel/src/org/eclipse/nebula/widgets/badgedlabel/BADGE_COLOR.java
lcaron: I prefer to let the use choose.

widgets/badgedlabel/org.eclipse.nebula.widgets.badgedlabel/src/org/eclipse/nebula/widgets/badgedlabel/BADGE_LOCATION.java
lcaron: Done. Enumeration removed and location is based on a combination of SWT.LEFT or SWT.RIGHT/SWT.TOP or SWT.BOTTOM

widgets/badgedlabel/org.eclipse.nebula.widgets.badgedlabel/src/org/eclipse/nebula/widgets/badgedlabel/BADGE_COLOR.java
wimjongman: If this breaks API then the major version number of the widget needs to be updated.

widgets/badgedlabel/org.eclipse.nebula.widgets.badgedlabel/src/org/eclipse/nebula/widgets/badgedlabel/BADGE_COLOR.java
wimjongman: Maybe the textColor could be something like BADGE_COLOR_AUTOMATIC for Mickael's idea to work.

lcaron: In fact I've created a set of predefined colors, but ones can specify foreground and background colors.
So I prefer to let developpers choose predefined color (blue, cyan...) or choose specific colors.

widgets/badgedlabel/org.eclipse.nebula.widgets.badgedlabel/src/org/eclipse/nebula/widgets/badgedlabel/BADGE_COLOR.java
lcaron: It does not break API because this widget has never been commited.
Breaking API is  called in my company the "docker" effect

wimjongman: I did not realize that. Thanks.

widgets/badgedlabel/org.eclipse.nebula.widgets.badgedlabel/src/org/eclipse/nebula/widgets/badgedlabel/BADGE_COLOR.java
wimjongman: Developers are not always in control of the colors. E.g. when theming is used.

lcaron: Good point for you !
I'll add the "automatic" badge text color this afternoon

PR #40, opened 2019-02-07
Bug 544177 - Support customizing regex for word boundaries

jfaltermeier:
This is a first draft. I added this to the RichTextViewerExample for testing. I am not sure about unit testing (I think currently there are none?) and about Since-Tags. 

Changes:
* Additional constructor for RichTextCellLabelProvider, RichTextPainter,
and TextPaintInstruction to pass in a regex
* Setter for RichTextPainter, and RichTextViewer to pass in a regex
* Adjust RichTextViewerExample to split after whitespace characters on
on "-"

### DIFF
--- a/widgets/richtext/org.eclipse.nebula.widgets.richtext.example/src/org/eclipse/nebula/widgets/richtext/example/RichTextViewerExample.java
+++ b/widgets/richtext/org.eclipse.nebula.widgets.richtext.example/src/org/eclipse/nebula/widgets/richtext/example/RichTextViewerExample.java
@@ -55,6 +55,7 @@ public class RichTextViewerExample {
 		GridDataFactory.fillDefaults().grab(true, true).applyTo(editor);
 
 		final RichTextViewer viewer = new RichTextViewer(parent, SWT.BORDER | SWT.WRAP);
+		viewer.setWordSplitRegex("\\s|\\-");//wrap after whitespace characters and delimiter
 		GridDataFactory.fillDefaults().grab(true, true).span(1, 2).applyTo(viewer);
 
 		final Text htmlOutput = new Text(parent,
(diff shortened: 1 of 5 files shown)


eclipsewebmaster:
Issue tracker reference:
https://bugs.eclipse.org/bugs/show_bug.cgi?id=544177

fipro78 (commented):
Why a constructor and a setter? To reduce complexity I would suggest to only have a setter.

jfaltermeier:
Thanks, I removed the additional constructor from RichtTextPainter. 
In TextPaintInstruction the regex is used in the constructor already, so we need to pass it there. 
I am not sure if a setter would work for the RichTextCellLabelProvider as this would require an update of the column's cells. Can we trigger this from the label provider itself in a nice way?

fipro78 (commented):
Another question just for understanding your problem. How would your regex look like? Just wondering if it doesn't make sense to simply extend the regex instead of making it configurable.

jfaltermeier:
As described in https://bugs.eclipse.org/bugs/show_bug.cgi?id=544177 the main motivation for us is the zero-width-space for which we would like to enable word wrapping. It could make sense to also add it to the default regex, however we then lose strict backward compatibility. 

A use case that comes to my mind that speaks for cutomizing the regex would be wrapping on special chars like "/"
Here you might want to break: "QWE/RTZ/UIO/PAS/DFG/HJK/LYX/VBT"
And here the current default would be fine: "Please send me the log at QWE/RTZ"

sdirix:
Hi @fipro78, do you have time to look at this change? As the change is backwards compatible and opt-in only there should be no difference for any of the existing users of RichText.

Please let us know if and how we shall adapt this PR.

fipro78:
@sdirix @jfaltermeier 
I just noticed that it was forgotten to increase the version number. As you added a new feature, the version for the richtext plugin and the feature should be 1.3. Could you please create a pull request that updates the version? Also remember to update the pom.xml files ... I always forget them and need to update the PR after the build failed. :)

jfaltermeier:
@fipro78 I opened up https://github.com/eclipse/nebula/pull/46 for this

PR #50, opened 2019-02-20
Bug 498010 - TableCombo w/image only : dropdown not displayed

lcaron:
When one clicks on the image, the drop down was not displayed.
Fix inspired by the patch made by Nicolas Cazottes


eclipsewebmaster:
Issue tracker reference:
https://bugs.eclipse.org/bugs/show_bug.cgi?id=498010

PR #54, opened 2019-02-21
Bug 458265

lcaron:
When item foreground is set to white, text is not readable on white
background
I fixed another problem : when you open a combo (table displayed) and you press ESC key, the popup is now closed.


eclipsewebmaster:
Issue tracker reference:
https://bugs.eclipse.org/bugs/show_bug.cgi?id=458265

PR #55, opened 2019-02-21
Bug 543044 - locale error

fipro78:
This fix will load the supported languages in ckeditor to be able to also support languages with additional country information.


eclipsewebmaster:
Issue tracker reference:
https://bugs.eclipse.org/bugs/show_bug.cgi?id=543044

PR #61, opened 2019-04-02
Bug 546069 - Filtering by Value reads special characters as regex

smolaro:
Signed-off-by: Stephen J. Molaro <stephen.j.molaro@boeing.com>


wimjongman:
jenkins add to whitelist

PR #69, opened 2019-04-17
Bug 543869 - PGroup inherits its foreground color to children widgets

jfaltermeier:
* Do not override group's foreground color from RectangleGroupStrategy
with the title foreground color
* Rather set the title foreground color on the gc when required and then
reset the group's foreground color


lcaron:
Seems ok for me

wimjongman:
add to whitelist

wimjongman:
Thanks Johannes

PR #77, opened 2019-04-28
Bug 289292 - Impossible to traverse to next field

lcaron:
Shift key is no longer filtered


wimjongman:
Thank you Laurent. You're on a roll!

PR #87, opened 2019-05-25
TextUtils performance optimization

uyong:
when text is large, continually substring opreration will produce much temporary string, then occur gc continually, cause app be at a standstill and loss of response


wimjongman:
Thanks for the patch @uyong. Eclipse requires you to sign the ECA before we can accept your patch.

1. Make an eclipse user account
2. Sign the ECA

https://www.eclipse.org/legal/ECA.php

uyong:
Thanks, i have signed the ECA, what next?

PR #88, opened 2019-05-27
Bug 547732: TextUtils performance optimization

uyong:
when text is large, continually substring opreration will produce much temporary string, then occur gc continually, cause app be at a standstill and loss of response.

Signed-off-by: uyong <gongguangyong@live.cn>

### DIFF
--- a/widgets/grid/org.eclipse.nebula.widgets.grid/src/org/eclipse/nebula/widgets/grid/internal/TextUtils.java
+++ b/widgets/grid/org.eclipse.nebula.widgets.grid/src/org/eclipse/nebula/widgets/grid/internal/TextUtils.java
@@ -18,6 +18,7 @@ import org.eclipse.swt.graphics.GC;
  * 
  * @author chris.gross@us.ibm.com
  * @author Mirko Paturzo <mirko.paturzo@exeura.eu>
+ * @author gongguangyong@live.cn
  *
  * Mirko modified the pivot calculation for improve short text provider performance. 
  * The pivot number is calculate starting from the size of the cell provided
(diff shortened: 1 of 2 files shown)


wimjongman:
ok to test.

The above is a Jenkins command.

wimjongman:
@uyong there is still some problem with the eca it seems. Is your e-mail address in Github the same as the one you signed your ECA for?

uyong:
yes, they are same.

wimjongman:
Bug 547732 - TextUtils performance optimization
https://bugs.eclipse.org/bugs/show_bug.cgi?id=547732

widgets/grid/org.eclipse.nebula.widgets.grid/src/org/eclipse/nebula/widgets/grid/internal/TextUtils.java
```diff
@@ -98,51 +57,46 @@ public static String getShortText(GC gc, String t, int width)
...
-        }
+	if (width >= gc.stringExtent(t).x) {
+	    return t;
+	}
+	
+	char[] chars = t.toCharArray();
+	int length = chars.length;
+	int left = 0;
+	int right = length - 1;
+	int calcWidth = gc.stringExtent("...").x;;
```
wimjongman: double semicolon ;;

wimjongman:
Thanks for the patch uyong.

I see that there is now only one method. However, the first method was using gc.stringExtent and the second routine was using gc.textExtent. They are different. Please see the javadoc GC. [1]

It is fine to remove the duplicate code but still each of them has to use stringExtent and textExtent respectively.

[1] https://help.eclipse.org/kepler/topic/org.eclipse.platform.doc.isv/reference/api/org/eclipse/swt/graphics/GC.html#stringExtent(java.lang.String)

uyong:
You are right,  i was noticed their different, but i think that the two methods keep only one is enough, there are two reasons:
1. the first method 'getShortText' that was using 'gc.textExtent', it have 0 reference in project of 'org.eclipse.nebula.widgets.grid' and  class of 'TextUtils' is internal.
2. most importantly, the first method 'getShortText' and the second method 'getShortString' their function is  very similar, i think of them as the same, so i use 'gc.getCharWidth' compute the width.

wimjongman:
I'm not happy with this approach. The class is effectively API so we cannot change it even if it is not used by us.

I suggest you keep the two original methods and mark them as deprecated pointing to your new method. Change all Grid classes that use the old methods and point them to the new one. You can mark the old methods for removal as well (in two years time.)

widgets/grid/org.eclipse.nebula.widgets.grid/src/org/eclipse/nebula/widgets/grid/internal/TextUtils.java
uyong: thank your proposal, i have fixed it.

wimjongman (approved):
Looks great. Thanks @uyong.

PR #107, opened 2019-09-03
Bug 550676 - the Checkbox button  also be possible on the MessageArea

sibeon:
The Checkbox feature should also be possible in the MessageArea.

Signed-off-by: Stefan Nöbauer <stefan.noebauer@kgu-consulting.com>


lcaron (approved):
Very good job, congratulations !

PR #169, opened 2019-10-01
Bug 551658 - preempt Java 9 deprecations

PeterC-DLS:
Migrate from using Number constructors and use static methods


widgets/visualization/org.eclipse.nebula.visualization.widgets/src/org/eclipse/nebula/visualization/widgets/figureparts/RoundScaleTickLabels.java
```diff
@@ -247,17 +247,16 @@ private boolean isMajorTick(double tickValue) {
      * 
      * @param base
      *            the base
-     * @param expornent
+     * @param exponent
```
lcaron, line 250: I like when typing error are fixed

lcaron (approved):
Pull request approved, you can merge it :)

PR #170, opened 2019-10-01
Bug 551670 - fix incorrect trace positioning

PeterC-DLS:
Ensure local range is set correctly


lcaron (approved):
Review done. Fine for me

PR #171, opened 2019-10-01
Bug 551671 - draw point style fixes

PeterC-DLS:
When a point style provider is given, do not ignore its point style when checking if it is none, and use its point size when drawing a filled circle.


lcaron (approved):
Review done. Fine for me

PR #176, opened 2019-10-04
Bug 551779 - LED Display Widget

lcaron:


fipro78:
Just a small question. Do you really want to still use EPLv1 or should we think about using EPLv2 nowadays?

lcaron:
> 
> 
> Just a small question. Do you really want to still use EPLv1 or should we think about using EPLv2 nowadays?

Sorry Dirk for my late answer,
Well I copied/pasted the licence information, and after digging all widgets are under EPLv2.
What do you think about opening a bug to migrate all components to EPLv2 ?
What are the impact of such a change ?
@wimjongman what is your opinion ?

fipro78:
I need to check with the foundation if we can simply update to EPLv2 or if we need some additional steps. Actually I think updating to EPLv2 should be the right direction.

lcaron:
> 
> 
> I need to check with the foundation if we can simply update to EPLv2 or if we need some additional steps. Actually I think updating to EPLv2 should be the right direction.

Ok thank you !

wimjongman:
We can just update the headers, see the FAQ. I will make a notice in the mailing list that we are switching.

https://www.eclipse.org/legal/epl-2.0/faq.php

PR #187, opened 2019-10-08
Bug 511353 - Support render link and manage link interaction in RichText

lcaron:
When a user add an hyperlink (anchor < a href="..." >) the link is underline, and a click on it opens the default browser with the URL.

### DIFF
--- a/widgets/richtext/org.eclipse.nebula.widgets.richtext.example/src/org/eclipse/nebula/widgets/richtext/example/RichTextViewerURLExample.java
+++ b/widgets/richtext/org.eclipse.nebula.widgets.richtext.example/src/org/eclipse/nebula/widgets/richtext/example/RichTextViewerURLExample.java
@@ -1,0 +1,57 @@
+/*****************************************************************************
+ *  Copyright (c) 2015, 2019 CEA LIST.
+ *
+ * All rights reserved. This program and the accompanying materials
+ * are made available under the terms of the Eclipse Public License v1.0
+ * which accompanies this distribution, and is available at
+ * http://www.eclipse.org/legal/epl-v10.html
+ *
+ * Contributors:
+ *		Laurent Caron <laurent.caron@gmail.com> - Initial API and implementation
+ *
+ *****************************************************************************/
+package org.eclipse.nebula.widgets.richtext.example;
+
+import org.eclipse.jface.layout.GridDataFactory;
+import org.eclipse.nebula.widgets.richtext.RichTextViewer;
+import org.eclipse.swt.SWT;
+import org.eclipse.swt.layout.FillLayout;
+import org.eclipse.swt.layout.GridLayout;
+import org.eclipse.swt.widgets.Composite;
+import org.eclipse.swt.widgets.Display;
+import org.eclipse.swt.widgets.Shell;
+
+public class RichTextViewerURLExample {
+
+	public static void main(String[] args) {
+		Display display = new Display();
+
+		final Shell shell = new Shell(display);
+		shell.setText("SWT RichTextViewer with URL");
+		shell.setSize(800, 600);
+
+		shell.setLayout(new GridLayout(1, true));
+
+		RichTextViewerURLExample example = new RichTextViewerURLExample();
+		example.createControls(shell);
+
+		shell.open();
+		while (!shell.isDisposed()) {
+			if (!display.readAndDispatch())
+				display.sleep();
+		}
+		display.dispose();
+	}
+
+	public void createControls(Composite parent) {
+		parent.setLayout(new FillLayout());
+
+		final RichTextViewer viewer = new RichTextViewer(parent, SWT.BORDER | SWT.WRAP);
+		viewer.setWordSplitRegex("\\s|\\-");// wrap after whitespace characters and delimiter
+		GridDataFactory.fillDefaults().grab(true, true).span(1, 2).applyTo(viewer);
+
+		viewer.setText("Please visite Nebula WebSite <a href=\"https://www.eclipse.org/nebula/\">Eclipse Nebula Web Site</a><br/>" + //
+				"You can also go to <a href=\"https://github.com/eclipse/nebula\">Github</a>");
+	}
+
+}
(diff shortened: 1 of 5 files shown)


fipro78 (commented):
I need to check your proposal in more detail. The TagProcessingState class for example is completely changed, so to see the real changes is not possible. Probably a line ending issue.
And why is there an anchor stack? Can anchors be wrapped with other anchors that would need to have a stack?

Personally I am not sure if the implementation covers all aspects. What if there is a line break? Does the area for the hover processing take care of that? And what if the link should be used for internal references? The current implementation only supports to directly open an external browser.

I actually thought of a more general "action item" implementation that can be extended e.g. for opening a browser or similar.

Please do not merge until I have found the time to review in more detail!

lcaron:
> I need to check your proposal in more detail. The TagProcessingState class for example is completely changed, so to see the real changes is not possible. Probably a line ending issue.

Sorry of that. In this class I just added the anchorStack + a push method + a poll method

> And why is there an anchor stack? Can anchors be wrapped with other anchors that would need to have a stack?
> 
Can you explain please ?

> Personally I am not sure if the implementation covers all aspects. What if there is a line break? Does the area for the hover processing take care of that? 

No I forgot this case. I'm gonna check it

> And what if the link should be used for internal references? The current implementation only supports to directly open an external browser.
> 

True, I just worked on the described use-case.

> I actually thought of a more general "action item" implementation that can be extended e.g. for opening a browser or similar.
> 

Good idea, even if I have no idea how to implement it : a list of action (first hyperlink=action and so on) ? An attribute ? Or a map URL/Action ? (for instance if href ="refresh" call the method refreshScreen() associated to this url) ?

> Please do not merge until I have found the time to review in more detail!

Of course :)
Thanks for your time.

fipro78:
> > I need to check your proposal in more detail. The TagProcessingState class for example is completely changed, so to see the real changes is not possible. Probably a line ending issue.
> 
> Sorry of that. In this class I just added the anchorStack + a push method + a poll method
> 
> > And why is there an anchor stack? Can anchors be wrapped with other anchors that would need to have a stack?
> 
> Can you explain please ?
> 

Well, I added the stack design to handle nested tags. For example the styling can be nested to first specify a background color and only for a single word a different foreground color. To handle those scenarios a stack needs to be used, as the styles need to be closed in the correct order by still supporting the inheritance of the style setting. Same for the font, if e.g. only a single word should be italic for a sentence in a specific font type.
I don't think such a stack behavior is needed for links or action items, because AFAIK links can not be nested with other links.

> > Personally I am not sure if the implementation covers all aspects. What if there is a line break? Does the area for the hover processing take care of that?
> 
> No I forgot this case. I'm gonna check it
> 

IMHO this is a complicated one. Because a text that can break should support the same action. But then a modified region needs to be set for the rest of the text on the new line.

> > And what if the link should be used for internal references? The current implementation only supports to directly open an external browser.
> 
> True, I just worked on the described use-case.
> 

For the described use case surely enough, and if the region can be determined for the link, this can be adapted later. Just wanted to point out which direction I would like to go here. One use case could be to scroll to a row in a NatTable similar to cross references in a table.

> > I actually thought of a more general "action item" implementation that can be extended e.g. for opening a browser or similar.
> 
> Good idea, even if I have no idea how to implement it : a list of action (first hyperlink=action and so on) ? An attribute ? Or a map URL/Action ? (for instance if href ="refresh" call the method refreshScreen() associated to this url) ?
> 

That actually needs to be defined. Maybe with a prefix in the href attribute that specifies the action. This would even match the HTML spec IIRC. We have a link and the URI defines the type or reference. 

> > Please do not merge until I have found the time to review in more detail!
> 
> Of course :)
> Thanks for your time.

Another possible issue is the integration in other widgets, e.g. JFace viewer and NatTable. Does the hover listener work correctly when integrating the RichTextViewer as viewer in a table?

lcaron:
Thank you for your feedback.
You're totally right, a stack for the Anchor Element is not relevant. I'll change it.

I've dug the problem of URL split into 2 lines (word wrap). I agree : it seems VERY complicated, but I have an idea. I'll try it.

For the configuration issue, I propose your idea : if the URL is http(s)://... the default browser will be opened. if the url is action://XXXX the action "XXXX" will be called.

Finally, I'm gonna check the "hover" possible issue with a viewer. Do you have an example of using RichTextViewer with another widgets ?

Thanx

fipro78:
There is a JFace viewer integration example here: https://github.com/eclipse/nebula/blob/master/widgets/richtext/org.eclipse.nebula.widgets.richtext.example/src/org/eclipse/nebula/widgets/richtext/example/JFaceViewerIntegrationExample.java

And for NatTable I have a dedicated example here:
https://github.com/eclipse/nebula.widgets.nattable/blob/master/org.eclipse.nebula.widgets.nattable.examples/src/org/eclipse/nebula/widgets/nattable/examples/_400_Configuration/_424_NebulaRichTextIntegrationExample.java

My fear is that your modification has a negative effect on the hover feature in NatTable. But I have not verified this. I will need to create an example that includes both.

lcaron:
> 
> 
> There is a JFace viewer integration example here: https://github.com/eclipse/nebula/blob/master/widgets/richtext/org.eclipse.nebula.widgets.richtext.example/src/org/eclipse/nebula/widgets/richtext/example/JFaceViewerIntegrationExample.java
> 
> And for NatTable I have a dedicated example here:
> https://github.com/eclipse/nebula.widgets.nattable/blob/master/org.eclipse.nebula.widgets.nattable.examples/src/org/eclipse/nebula/widgets/nattable/examples/_400_Configuration/_424_NebulaRichTextIntegrationExample.java
> 
> My fear is that your modification has a negative effect on the hover feature in NatTable. But I have not verified this. I will need to create an example that includes both.

I understand, 

Thank you for your time and support. After a sick leave I'm back to work, so I'll have less time.
Just to tell you : No pressure ;)

lcaron:
> Thank you for your time and support. After a sick leave I'm back to work, so I'll have less time.
> Just to tell you : No pressure ;)

Hi Dirk
Thank you for pointing the problems. You're totally right, this feature leads to unsolvable issues.
If you agree, I propose to mark this bug as "won't fix"

What do you think ?

fipro78:
Not sure if everything is unsolvable. But it definitely has a higher complexity and several side effects that need to be tested. From my point of view it can stay open and I will look after it when I have more time. At least the bugzilla ticket. 

You can close this pr if you think it is not the right direction.

lcaron:
Hi Dirk
Thank you for your feedback. Well I did not think it is unsolvable, but has you noticed it is much more complex than I expected and I think that my approach is going to a dead end.

I'm gonna close this PR but leave the bug open.

Have a nice evening,

Laurent

PR #189, opened 2019-10-10
Bug 552000 - [releng] Move to EPLV2

wimjongman:
Bug 552000 - [releng] Move to EPLV2

Fixed all java headers
Fixed all feature.properties
Fixed or added about.html to non-feature projects
Replaced epl-v10.html with epl-v2.0.html
replaced content of all license.html


wimjongman:
retest this please

PR #193, opened 2019-12-13
Bug 488927 - fix tick marks

PeterC-DLS:
When a fixed decimal precision format is used with sufficiently long axes, the grid step can become small enough so that neighbouring labels are duplicated. Duplicated ticks are removed but limited precision may round a tick value negative zero which then does not render the next label invisible. This fix ensures grid steps are big enough for the scale format to make tick labels distinct

### DIFF
--- a/widgets/visualization/org.eclipse.nebula.visualization.xygraph/src/org/eclipse/nebula/visualization/xygraph/linearscale/LinearScaleTicks.java
+++ b/widgets/visualization/org.eclipse.nebula.visualization.xygraph/src/org/eclipse/nebula/visualization/xygraph/linearscale/LinearScaleTicks.java
@@ -271,8 +271,23 @@ public class LinearScaleTicks implements ITicksProvider {
 		} else {
 			gridStep = Math.pow(10, exp); // 1*10^exponent
 		}
+
+		if (exp < 0) { // ensure grid steps are printable for scale format
+			String pattern = scale.getFormatPattern();
+			if (!pattern.contains("E")) {
+				int e = pattern.lastIndexOf(".") + 1;
+				if (e > 0) {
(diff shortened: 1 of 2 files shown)


widgets/visualization/org.eclipse.nebula.visualization.xygraph/src/org/eclipse/nebula/visualization/xygraph/linearscale/LinearScaleTicks.java
```diff
@@ -475,6 +490,11 @@ private void updateTickLabelForLinearScale(double min, double max, int length, d
 			tickLabelPositions.add(tickLabelPosition);
 		}
 
+		if (scale.isHorizontal()) {
+			System.err.println("For length of " + length);
```
lcaron: I do not know the logging philosophy in Nebula. Do you think this code is relevant ? Should it be removed ?

lcaron (approved):
Wondering if System.err.println() is relevant

widgets/visualization/org.eclipse.nebula.visualization.xygraph/src/org/eclipse/nebula/visualization/xygraph/linearscale/LinearScaleTicks.java
PeterC-DLS: No, accidentally left in debugging.

wimjongman:
Peter can this be merged?

PeterC-DLS:
Hi Wim,

Yes, I removed the debugging so it's a clean patch now.

Regards,
Peter

Get Outlook for Android<https://aka.ms/ghei36>

________________________________

PR #205, opened 2020-01-29
Bug 559648 - Convert XViewer column multi edit to use filtered dialog

dondunne:


wimjongman:
Thanks, Donald. XViewer has a great run. A very nice widget. Thanks for your support.

Issue #231, opened 2020-04-01
Notifier widget leaks one image per instance


ConradGroth:
For every Notifier instance there's one Image created (in line 211) which isn't disposed.
![image leak in notifier widget](https://user-images.githubusercontent.com/22416421/78144805-62057b00-7430-11ea-829e-712079c4f379.png)

To verify, import the org.eclipse.swt.tools.spies project and add the following lines to the NotifierSnippet:
```
public static void main(final String[] args) {
  DeviceData data = new DeviceData();
  data.tracking = true;
  final Display display = new Display(data);
  Sleak sleak = new Sleak();
  Shell sleakShell = new Shell();
...
```

ConradGroth:
See also https://bugs.eclipse.org/bugs/show_bug.cgi?id=561648

I cannot push my change to github (access denied). But the fix is easy, just add the following line after line 233:
`SWTGraphicUtil.addDisposer(shell, newImage);`

wimjongman:
Conrad let's get you set up. You need to fork the repository.

![image](https://user-images.githubusercontent.com/180969/78912035-6b719180-7a87-11ea-9107-20f359cc123f.png)


Then you can make a new branch in your clone and make a pull request. Let me know if you need more help.

Thanks!!

lcaron:
Patch merged today, thank you Conrad !

PR #233, opened 2020-04-17
Bug 561648 Notifier widget leaks one image per instance

ConradGroth:
Signed-off-by: Conrad Groth <info@conrad-groth.de>


ConradGroth:
This fixes #231

lcaron:
Patch merged in nebula.
Thank you very much Conrad for fixing this bug !

PR #236, opened 2020-06-29
Bug 564533 - NullPointerExeption when used with FormToolkit

lcaron:
Add null-safe tests to functions introduced when I developped the theming plugin.


wimjongman:
Thanks, Laurent!

PR #263, opened 2020-09-16
Updated target to photon

fipro78:
Signed-off-by: Dirk Fauth <dirk.fauth@googlemail.com>


wimjongman:
Thank you, brother! I was fighting this as well but could not see what was wrong.

fipro78:
Well we are not done yet. There are still issues when building with Java 11 and some tests that are failing sometimes on Jenkins. We need to tackle that once we want to update to Tycho 2.0 as that requires Java 11. But for now the builds are green again. :)

PR #264, opened 2020-09-27
Bug 567397 - ProgressCircle is jittery

wimjongman:
The animation is no longer ran in a separate job but is done directly.
There is a new method to set the animation delay.

Signed-off-by: Wim Jongman <wim.jongman@remainsoftware.com>


lcaron (approved):
Good job, ok for me

PR #287, opened 2020-11-26
Fix the clipping of the Grid on Linux GTK [Issue #286]

### LINKED ISSUE #286: Grid inside Scrollable Composite draws outside its bounds on Linux GTK
When a Grid is placed inside of a ScrollableComposite, then the grid contents are sometimes rendered outside of the area of the scrollable. This is most noticeable on the initial creation, and when resizing the entire window. Scrolling the grid stays within the bounds, but leaves a mess outside of the scrollable.

ScrollableGridTest-issue --- should be --> ScrollableGridTest-with-fix

_Test code that demonstrates this: [ScrollableGridTest.java.txt](https://github.com/eclipse/nebula/files/5598627/ScrollableGridTest.java.txt)_

Discovered on Linux with GTK (Ubuntu 20.04), and yet not a problem on Windows. This used to be fine on older SWT/GTK (Eclipse 3.8) and issue has surfaced since moving to a newer platform (i.e. Eclipse 4.1x). Perhaps something changed within GTK 3?

I think this may be due to the calls to **gc.setClipping()** in onPaint() before calling the paint() method of the cell renderer, as these are ignoring the clipping from the paint event.

I was able to fix it by saving the original clipping at the top of the function, and then replacing `gc.setClipping(new Rectangle(...))` with `gc.setClipping(originalClipping.intersection(new Rectangle(...)))`. This ensures that the clipping is never set to be outside that of the paint event. See [Grid.diff.txt](https://github.com/eclipse/nebula/files/5598703/Grid.diff.txt). Might need to do something with insertMarkPosFound block too.

MrTimbones:
This ensures that the rendering of each cell stays inside the bounds of the paint event, fixes #286

### DIFF
--- a/widgets/grid/org.eclipse.nebula.widgets.grid/src/org/eclipse/nebula/widgets/grid/Grid.java
+++ b/widgets/grid/org.eclipse.nebula.widgets.grid/src/org/eclipse/nebula/widgets/grid/Grid.java
@@ -5189,6 +5189,7 @@ public class Grid extends Canvas {
 		boolean insertMarkPosFound = false;
 
 		final GridCellSpanManager cellSpanManager = new GridCellSpanManager();
+		Rectangle originalClipping = e.gc.getClipping();
 
 		e.gc.setBackground(getBackground());
 		this.drawBackground(e.gc, 0, 0, getSize().x, getSize().y);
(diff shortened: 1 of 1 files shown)


wimjongman:
Thank you, Tim!!

Tim, I have to ask you to jump through some legal hoops to make your code EPL. Can you take a look?

https://github.com/eclipse/che/wiki/Eclipse-Contributor-Agreement

MrTimbones:
I've signed ECA and recommitted the change (first ever contribution to GitHub)

wimjongman:
Jenkins, add to the whitelist. Ok to test.

lcaron (approved):

PR #307, opened 2021-03-20
Fix a "Widget is Disposed" Exception in the Carousel widget

lcaron:
When playing with the carousel widget I discovered that a WidgetIsDisposed exception is thrown if one navigates too fast.


wimjongman:
Thanks, Laurent!

PR #324, opened 2021-04-06
Fix for Bug #527035

mpscholz:
This is a fix for Bug https://bugs.eclipse.org/bugs/show_bug.cgi?id=572035: "CDT.CLOCK_DISCRETE does not work with CDT.DROP_DOWN when also selection of the date"

The fix makes the DatePicker considering the flag CDT.CLOCK_DISCRETE now also in cases when the CDateTime is configured to display a date in drop down mode. In this case, a DicreteTimePicker is used; before, always an AnalogTimepicker was instantiated in such a case, regardless of the CDT.CLOCK_DISCRETE flag.

Note: tried to provide a fix that is as non-intrusive to the existing code as possible to reduce risk and to keep the change request small. Hence, the solution is working but not optimal. However, the correct approach (introduction of a time picker interface and a robust API for accessing the pickers also in drop down mode) would require a bunch of structural changes to the CDateTime and DatePicker classes.


lcaron:
Hi

First of all I would like to congratule you for the comment.
Unfortunately there is a conflit with the fiche DatePicker :'(

Can you please solve it and push again ?

mpscholz:
> Hi
> 
> First of all I would like to congratule you for the comment.
> Unfortunately there is a conflit with the fiche DatePicker :'(
> 
> Can you please solve it and push again ?

Wow, that was quick. :)
I notice that the conflicts are due to formatting.
However, as far as I know I have used the standard Eclipse formatter (in-built in Eclipe 2020-12) - is there a particular formatter for Nebula sources that must be used?

mpscholz:
The conflicts have been resolved. :)

lcaron:
Nice job, congrats ! I thank you very much for your work.

Can you please close the bug in Bugzilla ?

mpscholz:
I have changed the status of the Bug to CLOSED FIXED.

Maybe someone could change the assignee to "Matthias Paul Scholz", in case that is possible?

lcaron:
Don't you have the "Take" button next to the "Assignee" label in Bugzilla ?

mpscholz:
> Don't you have the "Take" button next to the "Assignee" label in Bugzilla ?

Unfortunately, I don't. :(
I presume that the availability of that button depends on some role of the person logged in (?).

lcaron:
Matthias, I've assigned the bug to you. Thank you !

PR #325, opened 2021-04-08
Bug: - Align formatting of the Notifier sources with the actual formatting standard

mpscholz:
Prior to this change, the sources of the notifier plugins were formatted in a format different from the one the actual Eclipse IDE formatter uses.
This would result in a lot of additional format changes when submitting a new pull request; hence, it would be hard for a reviewer to sort out the real funtional changes in the request.

This change reformats the source code of the notifier plugins using the formatter built-in into the Eclipse version 2020-12 to get a clean copy for further pull requests.
It does not contain any functional change whatsoever.

Signed-off-by: Matthias Paul Scholz <matthias.paul.scholz@gmail.com>


wimjongman:
add to whitelist

lcaron (approved):
Thank you !

PR #326, opened 2021-04-10
Bug 572752 - Update ViewSelectedCellDataAction to support IXViewerPre…

dondunne:
…ComputedColumn.  Copy/View selected cell data doesn't handle the newer IXViewerPreComputedColumn.  Update to handle.


dondunne:
Abandon

PR #332, opened 2021-05-11
Fix #331 DualList.select(int) do not work as expected

### LINKED ISSUE #331: DualList.select(int) do not work as exspected
If I call `DualList.select(int)` the item is present in both lists

laeubi:
Signed-off-by: Christoph Läubrich <laeubi@laeubi-soft.de>


laeubi:
@wimjongman can you review this?

PR #369, opened 2021-09-19
Bug 368 - How to change PTProperty color when the property changed?

lcaron:
Add the methods changeFont(), changeForegroundColor() and changeBackgroundColor() to the `PTProperty` Object
Add the discardChanges() method to the `PropertyTable `object


lcaron:
retest this please

PR #382, opened 2021-11-03
mvn clean verify fails on the root project with DuplicateProjectException

### LINKED ISSUE #378: mvn clean verify fails on the root project with DuplicateProjectException
If I run mvn clean verify in the root of the project I get:

org.eclipse.nebula.examples is duplicated in the reactor.

vogella:
Examples shall only be included once in the build. This change allows to
run "mvn clean verify" or "mvn verfify" on the root directory.

Fixes #378

### DIFF
--- a/releng/org.eclipse.nebula.nebula-incubation/pom.xml
+++ b/releng/org.eclipse.nebula.nebula-incubation/pom.xml
@@ -36,8 +36,6 @@ Contributors:
 	<modules>
 		<!-- Dependencies -->
 		<module>../org.eclipse.nebula.nebula-parent</module>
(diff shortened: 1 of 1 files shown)


wimjongman:
There are two products in the build. The "released widgets" and the "incubation widgets". They are all in the same repo. Both share the examples view because it is dynamic. We never build from the root but from the respective released or incubation parent.

I see that both builds are now failing: https://ci.eclipse.org/nebula/

vogella:
Please revert in this case

Wim Jongman ***@***.***> schrieb am Mi., 3. Nov. 2021, 17:42:

vogella:
https://github.com/eclipse/nebula/pull/389 should contain a better fix.

PR #394, opened 2021-11-05
Using lambdas in AbstractVTestCase

vogella:
With the JDT clean actions I converted the anonymous inner classes into
lambda expressions for AbstractVTestCase.


wimjongman:
Thanks!

PR #395, opened 2021-11-05
Using lambdas in SvgRenderTests

vogella:
With the JDT clean actions I converted the anonymous inner classes into
lambda expressions for SvgRenderTests. This results in more modern code
and removes some SonarLint warnings.


wimjongman:
Thanks!

PR #397, opened 2021-11-05
SVGRendererTests do not block the command line build anymore

### LINKED ISSUE #390: SVG tests stop local buid
If I run a local build for the released components, the test for the SVG view stops the build. Seems like the SVG tests not never stopped and the build waits for them.

vogella:
With this change, we can run the tests in the Maven build as they are
not  blocking anymore.


wimjongman:
Thanks, Lars!

PR #418, opened 2021-11-10
Allow that the widgets in PreferenceWindow can increase in size

### LINKED ISSUE #417: Allow that the widgets in PreferenceWindow can increase in size
Our RCP allows the user to use CSS to set a very large font. Unfortunately in the current implementation the widgets of the PreferenceWindow are not allow to grow. Would be great if this could be adjusted.

PR coming.

vogella:
Instead of setting the widthHint and heightHint we define the minimum.

Fixes #417


lcaron (approved):

lcaron:
PR Merged, thank you @vogella ;)

PR #431, opened 2021-12-02
Notify listeners even in case newFocus is not a VControl

### LINKED ISSUE #430: CDateTime does not notify focusLost
A FocusListener attached to CDateTime is not notified about a focusLost when the focus changes from the CDateTime to a non-VControl widget. Apparently the VTracker.setFocusControl(VControl control) ignores the possibility that control (which is received by VTracker.getVControl(...) in the FocusIn case in the VTracker.filter) may have been a standard SWT widget instead of a VControl-based widget.
I assume this bug was introduced in https://github.com/eclipse/nebula/commit/b9129cd66a3cf3fb7380cc4b996e354b95965016. Thus, the fix should especially not re-introduce the old bug https://bugs.eclipse.org/bugs/show_bug.cgi?id=476035.

The attached sample application [CDateTimeFocus.zip](https://github.com/eclipse/nebula/files/7641727/CDateTimeFocus.zip) demonstrates that error. Test steps:
- Focus the left CDateTime widget
- Press the "Any button" -> on standard out no focus lost message is printed
- Focus the left CDateTime widget again
- Focus the right CDateTime widget -> on standard out a focus lost message is printed

mikoeh:


mikoeh:
I have accidentaly commited as user mko while my github account runs with the user mikoeh as which I have signed the ECA.

wimjongman:
@mikoeh, you have to sign the ECA before we can accept patches. Make sure to use an Eclipse account with the same email as your GitHub account.

https://www.eclipse.org/legal/ECA.php

wimjongman:
There is a project with CDateTime tests. Can you run this?

mikoeh:
I ran the tests in
- org.eclipse.nebula.widgets.cdatetime.tests
- org.eclipse.nebula.cwt.tests

Both had test failures before my change and they are 100% identical after my change.

wimjongman:
> Can one of the admins verify this patch?

Ok to test

wimjongman:
Thank you @mikoeh! Congratulations on your first Nebula patch. We are looking forward to seeing more of you.

PR #435, opened 2021-12-30
Bug 433 - Modifylistener called a lot

lcaron:
Rollback for fix of issue #301


fipro78:
Removing the onchange plugin fixes the regression as far as I can see. I raised a question at the ckeditor community to hopefully get a hint on the issue.

lcaron:
Thank you @fipro78 for your feedback, the PR is merged. I keep working on the original issue with the modify listener

Issue #419, opened 2021-11-11
TextAssist appears in the background


Bauer-David:
I noticed that the popup of TextAssist is sometimes displayed in the background. The style is set to SWT.ON_TOP. This error occurs very sporadically and only under windows.

![Popup in the background](https://user-images.githubusercontent.com/20420949/140722328-a586ae4b-df5b-4758-9225-9072031de791.png)

In the picture above you can see how the popup is displayed in the background. If it helps, here is my line of code where I create the Text Assist.

`TextAssist text = new TextAssist(composite, SWT.BORDER, contentProvider);`

wimjongman:
Thanks, David. Can you supply a patch?

Bauer-David:
No, sorry. I was hoping that you might come to a solution. I'm still trying to solve it, but unfortunately I'm at a dead end.

wimjongman:
Before popup.setVisible(true) please set popup.setActive()

This will bring the popup to the foreground before it is made visible. Please try with this setting and if it works you can supply a pull request.

wimjongman:
That has glitches on windows. Try this block instead:

			if (!popup.isVisible()) {
				popup.setActive();
				popup.setVisible(true);
			}

Bauer-David:
Thank you. I will test your solution and give you feedback when I see results. It might take a while because the error is very sporadic. I will create a pull request if it works.

lcaron:
@Bauer-David Any news about this bug please ?

Bauer-David:
Sorry for the late reply. We have tested your solution but it was not successful. If we find another solution I will post it here

lcaron:
Ok, feel free to ask help !

Bauer-David:
We continued testing and tried a few solutions, but nothing fixed the error. We also noticed that the error occurs almost exclusively on native Windows. For example, it does not occur on Parallels. If you have any further solutions or help here, we would be very grateful. 

If it helps you, here is some more information about our application:

Used Java Version: Amazon Corretto 11 [11.0.6]
Used Framework: e4 RCP

Code of our time field:



		TextAssistContentProvider contentProvider = new TextAssistContentProvider() {
			@Override
			public List<String> getContent(String entry) {
				ArrayList<String> result = new ArrayList<>();
				Instant time = TimeUtil.getTime(entry);
				if (time == null && !entry.isEmpty()) {
					result.add(translationService.translate("@msg.ErrorConverting", null));
				} else {
					Preferences preferences = 
[...]

PR #451, opened 2022-02-25
#419: move popup above application

wis36:
solves #419


wis36:
didn't work because of a wrong email address

PR #452, opened 2022-02-28
#419: move popup above application

wis36:
the same with the right email address


lcaron:
PR merged.
Thank you @wis36 !

PR #484, opened 2022-08-03
Bug 344 - Image rendering for richtext editor

x21L:
It is now possible to insert images in the richtext editor.


fipro78:
Thanks for the contribution. I will have a look at this. But could you please avoid the code style changes? They make it really hard to focus on the real changes.

fipro78 (changes requested):
There is a reason why the default toolbar configuration is in the Java code and not the Javascript file. It provides the option that a user can change the configuration programmatically via setOptions(). And additionally the default configuration can be viewed and adapted in Java Code without the need to search for the right Javascript. And you changed the default configuration and extended it heavily. There are much more features now enabled than before, and several of those features are not even supported by the painter (e.g. Subscript / Superscript).
IMHO the toolbar configuration change needs to be reverted.

Regarding the rendering of the image, that also needs some rework.
a) it produces a resource leak as you simply create a new Image but you never dispose it. So after some rendering operations you will get a an error that no more handles are available.
b) it is a very simple adaption and IMHO does not cover all aspects of adding an image to a HTML. It simply draws the image without taking care about additional configurations (how should the text around the image be handled?), you are using incorrect parameters (drawImage(image, x, y) uses x and y as coordinates where to draw, but you are using height and width).

As already commented, you should avoid code style changes, as it makes it really hard to spot the real changes. And it would be good if you could squash your commits to a single commit. Makes it also easier to review and keeps the history of the project clean.

x21L:
Regarding the change of the JavaScript: do you know how it is possible to add the image adding functionality to the editor? If I enable it in Java, then I cannot configure it. There are a bunch of elements in the insert category which are not needed, insert flash for example. Insert is not a group, you must configure the items.

About the formatting: is there some formatting style to import available? I have seen your installer, but it does not seem to work, unfortunately.

Issue #515, opened 2023-05-23
Move FormattedText from incubation to release


chrhem:
Would it be possible to release FormattedText officially?
We have been using it in a commercial application since 2008 and have not had any problems with it.

Just because it is still in incubation, we mirror the incubation update site to achieve a stable build process.

I could make a PR for moving it into release.

@vogella

vogella:
@lcaron @wimjongman  WDYT?

lcaron:
I've no problem with that, if it fits your requirement.

@chrhem I'll be happy if you provide a patch. If you need help feel free to ask !

wimjongman:
+1

vogella:
Btw. why do we not move all incubator widgets to the regular build and adjust their manifest so that these widgets are listed as provitional API? This way the build could be simplified and the consumer of the non-released API could still see that the widget is not officially release.

wimjongman:
Because the incubator widgets have no maintainers, some of them are alpha, so it will cause more stress for the team.

vogella:
> Because the incubator widgets have no maintainers, some of them are alpha, so it will cause more stress for the team.

+1

lcaron:
> Because the incubator widgets have no maintainers, some of them are alpha, so it will cause more stress for the team.

+++1

marcolopes:
Hi! Been using FormattedText *since 2008* also!!

The only need i had was to *extend FormattedText *and implement a 
"*dispose*" method to "*detach*" the formatter and _*remove the 
listeners*_ (so the TEXT could be used on another *FormattedText* 
without disposing it)

and there is a small bug with *IntegerFormatter *since the first 
versions (also present in the latest 3.0.0 release) that causes this 
stacktrace when "deleting" the LAST CHAR of the TEXT

java.lang.StringIndexOutOfBoundsException: String index out of range: -1
     at java.lang.String.substring(String.java:1875)
     at 
org.eclipse.jface.internal.databinding.swt.SWTVetoableValueDecorator$1.handleEvent(SWTVetoableValueDecorator.java:38)
     at org.eclipse.swt.widgets.EventTable.sendEvent(EventTable.java:84)
     at org.eclipse.swt.widgets.Widget.sendEvent(Widget.java:1053)
     at org.eclipse.swt.widgets.Widget.sendEvent(Widget.java:1077)
     at org.eclipse.swt.widgets.Widget.sendEvent(Widget.java:1062)
[...]

PR #519, opened 2023-05-24
Move FormattedText from incubation to release #515

chrhem:
Needed to add FormattedText widget to incubation reactor build because DateChooser has a dependency to it

### DIFF
--- a/releng/org.eclipse.nebula.feature/feature.xml
+++ b/releng/org.eclipse.nebula.feature/feature.xml
@@ -254,4 +254,7 @@
          id="org.eclipse.nebula.widgets.roundedcheckbox.feature"
          version="0.0.0"/>
 
+   <includes
+         id="org.eclipse.nebula.widgets.formattedtext.feature"
+         version="0.0.0"/>
 </feature>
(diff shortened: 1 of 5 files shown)


releng/org.eclipse.nebula.nebula-incubation/pom.xml
```diff
@@ -38,10 +38,10 @@ Contributors:
 		<module>../org.eclipse.nebula.nebula-parent</module>
 		<module>../../examples/org.eclipse.nebula.examples</module>
 		<module>../../examples/org.eclipse.nebula.examples.feature</module>
+		<module>../../widgets/formattedtext</module>
```
wimjongman: Please move the line back to the original location.

wimjongman (changes requested):
Thanks for the patch @chrhem. 

I think there are some category.xml that sites also have to be changed. Can you take a look at that?

chrhem:
Hi @wimjongman , thanks for the quick review.
category.xml's are updated. PDE tooling changed some whitespaces. If this is unwanted i could revert that.
The line in the incubation pom.xml is back in it's old location.

wimjongman (approved):

wimjongman:
Thank you!! @chrhem

Issue #527, opened 2023-09-22
Oomph the build


wimjongman:
We want our builds to mimic what has been done for other projects, e.g. BIRT:

https://download.eclipse.org/birt/updates/index.html

Ed @merks, can you help us with this? I can do the lifting if you tell me what you need:

- [ ] Target platform
- [ ] ...

merks:
I see there is already a setup.  I've not tried it yet, but I will...  

Are you asking that you'd like the *.target to be generated as side effect of resolving the targlets as is current done for BIRT?

The build promotion stuff I understand.  I see your ci instances don't use a Jenkinsfile.  That would make configuration of the job much easier, and managed in the clone rather than separately.

Let me investigate how the setup current works and sketch out some scaffolding for what would be needed.  Both EMF and Oomph use too, so I have a vested interest in in be well maintained...

merks:
The setup doesn't work, so I started to repair it:
```
  ERROR: org.eclipse.equinox.p2.director code=10053 Cannot complete the install because one or more required items could not be found.
    ERROR: org.eclipse.equinox.p2.director code=0 Software being installed: artificial_root 1.0.0.v1695451715693
    ERROR: org.eclipse.equinox.p2.director code=0 Missing requirement: Draw2d 3.10.100.201606061308 (org.eclipse.draw2d 3.10.100.201606061308) requires 'java.package; com.ibm.icu.text 3.8.1' but it could not be found
```
If I import all the projects, there are many errors. 

![image](https://github.com/eclipse/nebula/assets/208716/328bc87d-3144-4843-b152-7ca5d488d0cc)

I guess I should not do that, but do they exist if they are not used/maintained?

Some I could try to fix, but I don't know data binding well to know what the replacements should be...

It's better if I only import the feature and its dependencies, but even then, there are errors:

![image](https://github.com/eclipse/nebula/assets/208716/b8df73ec-3656-4a30-ba72-7e1a28df20f6)

I've not looked at the build to see how the dependencies are specified, but I don't find a meaningful *.target...

I don't get the sense that the builds are producing things that are tested to work with (or even compiled against) the latest Eclipse Platform...

wimjongman:
Yes, many widgets are without an owner so the build can be stale for some projects when moving to newer Eclipses. I will take a look at that. I know the bidi stuff is special, @tomsontom should be able to comment on that.

tomsontom:
What do you mean with bidi? Eclipse Databinding?

[...]

PR #528, opened 2023-09-24
Ensure that the Oomph setup works

merks:
Add a setup configuration for an easier setup experience.

Fix FormattedTextObservableValue to use DisplayRealm.getRealm instead of SWTObservables.getRealm.


merks:
@wimjongman 

Action is required...

wimjongman:
Sorry, I missed this, Ed.

Issue #516, opened 2023-05-23
Use latest Tycho and Java 17 for building


vogella:
The current build does not work with Java 17 locally (Tycho to old), would it be OK to update the GH to Java 17 and Tycho to the latest 3.0.4 release (which requires Java 17)?

I don't have write access to the Nebula Jenkins, so if Java runtime changes are requried on Jenkins, you would have to do it.

cc @wimjongman @lcaron

merks:
Done via https://github.com/eclipse/nebula/issues/559

PR #543, opened 2023-12-23
Bug #516 - Use latest Tycho and Java 17 for building

lcaron:
Move Tycho version (and move to JDK 17)

### DIFF
--- a/.mvn/extensions.xml
+++ b/.mvn/extensions.xml
@@ -2,6 +2,6 @@
   <extension>
     <groupId>org.eclipse.tycho.extras</groupId>
     <artifactId>tycho-pomless</artifactId>
-    <version>2.5.0</version>
+    <version>3.0.4</version>
   </extension>
 </extensions>
--- a/releng/org.eclipse.nebula.nebula-parent/pom.xml
+++ b/releng/org.eclipse.nebula.nebula-parent/pom.xml
@@ -22,7 +22,7 @@ Contributors:
 
 	<properties>
 
-		<tycho-version>2.6.0</tycho-version>
+		<tycho-version>3.0.4</tycho-version>
 		<tycho-extras-version>${tycho-version}</tycho-extras-version>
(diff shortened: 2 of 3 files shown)


.mvn/extensions.xml
```diff
@@ -2,6 +2,6 @@
   <extension>
     <groupId>org.eclipse.tycho.extras</groupId>
     <artifactId>tycho-pomless</artifactId>
-    <version>2.5.0</version>
+    <version>3.0.4</version>
```
laeubi, line 5:
```suggestion
    <version>4.0.4</version>
```
The latest Tycho version as of today is [4.0.4](https://github.com/eclipse-tycho/tycho/releases/tag/tycho-4.0.4)

releng/org.eclipse.nebula.nebula-parent/pom.xml
```diff
@@ -22,7 +22,7 @@ Contributors:
 
 	<properties>
 
-		<tycho-version>2.6.0</tycho-version>
+		<tycho-version>3.0.4</tycho-version>
```
laeubi, line 25: You might want to use a `maven.config` file and then share the Tycho version for extension/pom in that file like described here: https://tycho.eclipseprojects.io/doc/latest/StructuredBuild.html#pomless-builds

laeubi:
There is no `<tycho-extras-version>` anymore for a long time so you probably want to replace that with search+replace

Boarschti42:
Hi, i recommend using 
```
<extensions>
    <extension>
        <groupId>org.eclipse.tycho</groupId>
        <artifactId>tycho-build</artifactId>
        <version>4.0.4</version>
    </extension>
</extensions>
```

instead of tycho-pomless since this is the current way suggested.

merks:
This is obsolete with the changes from https://github.com/eclipse/nebula/issues/559

PR #544, opened 2023-12-23
Organize the subset of imported projects into working sets

merks:


lcaron:
Thank you Ed for your work !

PR #546, opened 2023-12-26
Include incubation feature in setup

merks:
Use DisplayRealm.getReal rather than SWTObservables.getRealm. Accommodate generates in generics in GEF.

### DIFF
--- a/widgets/datechooser/org.eclipse.nebula.widgets.datechooser/src/org/eclipse/nebula/widgets/datechooser/DateChooserObservableValue.java
+++ b/widgets/datechooser/org.eclipse.nebula.widgets.datechooser/src/org/eclipse/nebula/widgets/datechooser/DateChooserObservableValue.java
@@ -5,7 +5,6 @@ import java.util.Date;
 import org.eclipse.core.databinding.observable.Diffs;
 import org.eclipse.core.databinding.observable.Realm;
 import org.eclipse.core.databinding.observable.value.AbstractObservableValue;
-import org.eclipse.jface.databinding.swt.SWTObservables;
 import org.eclipse.swt.SWT;
 import org.eclipse.swt.events.DisposeEvent;
 import org.eclipse.swt.events.DisposeListener;
@@ -63,7 +62,7 @@ public class DateChooserObservableValue extends AbstractObservableValue {
 	};
 
 	public DateChooserObservableValue(final DateChooser dateChooser, int updateEventType) {
-		this(SWTObservables.getRealm(dateChooser.getDisplay()), dateChooser, updateEventType);
+		this(Realm.getDefault(), dateChooser, updateEventType);
 	}
 
 	public DateChooserObservableValue(final Realm realm, final DateChooser dateChooser, int updateEventType) {
--- a/oomph.setup
+++ b/oomph.setup
@@ -123,14 +123,20 @@
     <description>${scope.project.label}</description>
   </setupTask>
   <setupTask
+      xsi:type="setup:StringSubstitutionTask"
+      name="nebula.git.clone.location"
+      value="${git.clone.location}"/>
+  <setupTask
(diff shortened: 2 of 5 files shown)


merks:
@lcaron 

The build are failing like for this PR for me locally as well. I think it needs version changes like this:

image

What I hope to get working is a local launch configuration for the Tycho build like this:

image

If I get that working, I can better determine which projects are build built and should be in the workspace and then eventually automating the publishing of update sites like I've done for other projects, most recently for GEF Classic...

widgets/datechooser/org.eclipse.nebula.widgets.datechooser/src/org/eclipse/nebula/widgets/datechooser/DateChooserObservableValue.java
```diff
@@ -63,7 +63,7 @@ public void widgetDisposed(DisposeEvent e) {
 	};
 
 	public DateChooserObservableValue(final DateChooser dateChooser, int updateEventType) {
-		this(SWTObservables.getRealm(dateChooser.getDisplay()), dateChooser, updateEventType);
+		this(DisplayRealm.getRealm(dateChooser.getDisplay()), dateChooser, updateEventType);
```
laeubi:
```suggestion
		this(Realm.getCurrent(), dateChooser, updateEventType);
```
at least that's what is used in the DatabindingFramework internally when no realm is passed.

merks: Did you mean Realm.getDefault()?  I don't claim to understand the subtleties of this, so choose to replace it with something that appears to be 100% semantically equivalent with what was there before.  E.g., it looks like getDefault works only on the display thread, but DisplayRealm.getRealm works on any thread, I think.  Of course, I don't know whether that is relevant here either...

laeubi: Yes `Realm.getDefault()` if you search for references you will se that this is always used when no explicit realm is passed in e.g. you can even look at the AbstractObservableValue default constructor that is extended here and I think its better to stick to this convention than trying to be "smart" here, `Realm.getDefault()` works on any thread that is currently a realm thread and one should only create observables from such thread or pass an explicit realm.

merks: I see what you mean, e.g., 

image

merks: There was another place using that "pattern" so I changed it too.

lcaron:
@merks I've fixed the build. Can you please rebase and keep me informed ?

merks:
FYI, I'm out of office until Friday.  Hopefully I'll have time then.

merks:
I think I have some mistaken conceptions about what should be in the target platform...  I can't look at this further until I'm back.

merks:
Okay, last try for today.  I was able to do both builds locally now...

merks:
@lcaron  @wimjongman 

This can be merged and then I can look again at how we might improve the production of update sites...

wimjongman:
Wonderful, thanks Ed.

wimjongman:
Is it easier for you if we make you a committer?

merks:
Yes, that would allow me to prototype everything without bothering you folks nearly as much...

lcaron:
Thank you Ed. You deserve to be a commiter. I'm gonna submit the request.

Issue #559, opened 2024-01-26
Improve releng


merks:
I wanted to establish (agree on) the goals for improving the releng for Nebula.

I assume one goals is to improve the update site structure as was done for other projects such as GEF recently:

https://download.eclipse.org/justj/?file=nebula

Do we want to keep maintaining the distinction between the non-incubating versus non-incubating widgets?  If we are to maintain two updates sites, I would prefer each to be rooted in a different child folder of the nebula folder, but with what names?

Another goal is to have the build driven by a multi-branch Jenkinsfile.  Probably we can do that with a single job even if we want to produce two separate update sites; I think it would still make sense to produce releases for the incubating widget...

@lcaron @wimjongman 

Please share your thoughts about what you would like to see as an end-result.

laeubi:
As a (random) user of nebula I never cared if something is "incubating" or not ... just put everything in one site, if there is a bug it must be fixed, at best by the reporting user if its really critical.

merks:
It's certainly easier to maintain one site and it's an interesting and useful data point that as a consumer, a single site is also more convenient.  Not only that, a bunch of incubating components have 1.x versions that suggest they are not incubating:

![image](https://github.com/eclipse/nebula/assets/208716/7c089717-20df-414a-a712-873680c149b0)

The builds don't appear to be using jgit stamping for the qualifier.  I wonder if another goal would be to change that?

wimjongman:
The incubation widgets are not stable. They should be separate or source only. In terms of root naming, we use **_incubation_** and  **_release_**.

laeubi:
@wimjongman does it really matter? I even use the SNPASHOTs an never get any "instability" given that there are not much contributions to nebula one should decide how much LTS support is available for a project.

e.g. for Tycho we always do MAJOR release now, so we don't promise 100% compatibility even though its often only a matter of increase the version number for 99% of the builds.

merks:
> The incubation widgets are not stable. They should be separate or source only. In terms of root naming, we use **_incubation_** and **_release_**.

Keep in mind the structure of an update site:

https://download.eclipse.org/justj/?file=tools/gef/classic

![image](https://github.com/eclipse/nebula/assets/208716/3b7f055c-7607-41dd-9f41-bca25b72da65)

The red things are legacy things before before JustJ was used for publishing.

[...]

PR #568, opened 2024-02-09
Improve and simplify the releng infrastructure

merks:
https://github.com/eclipse/nebula/issues/559


merks:
I don't know how to get these working.  The build uses Tycho 4.0.4 so it can't work with Java 11.   Also, the java 17 build fails because it needs Maven 3.9.0 which it apparently doesn't have...

PR #574, opened 2024-03-15
Add support to the slider to format the value

laeubi:
Currently the slider always show the raw number, but often this is not very suitable  for the user e.g. if the Number represents a unit of measure or should be formated with a number format.

This adds a new method NebulaSlider#setLabelFormatProvider that allows to customize the formatting of the raw number.

This then maybe looks like this:
grafik


lcaron:
Good job, thank you Christoph for this PR

PR #575, opened 2024-03-16
Only fire a selection event when the value changes

laeubi:
Currently the NebulaSlider fires a selection event on each move even if the value has not change numerically.

This first checks if the value has actually changed before redraw/event firing.


lcaron:
Thank you Christoph

PR #576, opened 2024-03-16
Add support for the default selection to the slider

laeubi:
Currently the slider fires an update event on each change, but this can be not desired e.g because changing the value might be an expensive operation, or even the user chooses the same value in the end.

This now captures the current value on mouse down and if it has changed on mouseup fires a default selection event. This way a user of the slider can always decide what approach works best or even combine the different approaches. For example one can use the fast changing value to update ui-elements and the default selection to persist the value to disk/database/preferences/...


laeubi:
@lcaron what do you think?

lcaron:
Fine for me, thank you very much @laeubi

PR #577, opened 2024-03-18
Support disabled state of slider

laeubi:
Currently one can make the slider disabled but it then shows as before and one can simply no longer move it, so the user is unable to see that this control is different.

This now adds a new disabled state where all colors are simply converted to gray.

## enabled

grafik

## disabled
grafik


lcaron:
Nice job, congrats and thank you.

PR #579, opened 2024-03-22
Cache the font of the slider / do not dispose colors

laeubi:
Currently the font and colors are created on each call of getTextFont / get...Color and as these methods are called on each render it creates many fonts/colors and their dispose listener.

This caches the font and removes the disposal of colors as SWT no longer acquire native resources for colors. Beside that a new getFontSize method is provided so one can easily override that.


lcaron:
PR Merged, thank you @laeubi

PR #580, opened 2024-03-22
Fix the Slider#computeSize method to take the real height into account

laeubi:
Currently the slider is displayed quite big compared to other SWT controls (e.g. labels) and even if one make the slider height smaller this does not really change much using the usual SWT layout managers. The reason for this is that the slider currently reports a minimum height of 40 instead of query the real selector height.

This now reports the real height the control will be painted including a one pixel padding on top/bottom for antialiasing effects to be correctly shown.


lcaron:
Nice contribution, thank you @laeubi

lcaron (approved):

PR #582, opened 2024-03-22
Rework SplitButton to not use SWT's internal TypedListener

HannesWell:
The `org.eclipse.swt.widgets.TypedListener` should be considered an internal class (as the doc states) and leaks the class `SWTEventListener` which also resides in an internal package.

This reworks `SplitButton` to not use `TypedListener` anymore.

### DIFF
--- a/widgets/splitbutton/org.eclipse.nebula.widgets.splitbutton/src/org/eclipse/nebula/widgets/splitbutton/SplitButton.java
+++ b/widgets/splitbutton/org.eclipse.nebula.widgets.splitbutton/src/org/eclipse/nebula/widgets/splitbutton/SplitButton.java
@@ -15,9 +15,9 @@
  *******************************************************************************/
 package org.eclipse.nebula.widgets.splitbutton;
 
-import java.util.Iterator;
 import java.util.LinkedList;
 import java.util.List;
+import java.util.function.Supplier;
 
 import org.eclipse.swt.SWT;
 import org.eclipse.swt.SWTException;
@@ -33,7 +33,6 @@ import org.eclipse.swt.widgets.Composite;
 import org.eclipse.swt.widgets.Event;
 import org.eclipse.swt.widgets.Listener;
 import org.eclipse.swt.widgets.Menu;
-import org.eclipse.swt.widgets.TypedListener;
 import org.eclipse.swt.widgets.Widget;
 
 /**
@@ -57,7 +56,8 @@ public class SplitButton extends Button {
 
 	private static final int EXTRA_WIDTH = 60;
 
-	private final List<Listener> listeners = new LinkedList<>();
+	private final List<Listener> selectionListeners = new LinkedList<>();
+	private final List<SelectionListener> typedSelectionListeners = new LinkedList<>();
 
 	private int splitButtonAreaLeft = -1;
 	private int splitButtonAreaTop = -1;
@@ -191,12 +191,15 @@ public class SplitButton extends Button {
 					getMenu().setVisible(true);
 
 				} else {
-					for (final Listener listener : listeners) {
-						final Event evt = new Event();
-						evt.widget = SplitButton.this;
-						evt.display = getDisplay();
-						evt.type = SWT.Selection;
+					Event evt = new Event();
+					evt.widget = SplitButton.this;
+					evt.display = getDisplay();
+					evt.type = SWT.Selection;
+					for (Listener listener : selectionListeners) {
 						listener.handleEvent(evt);
+					}
+					for (SelectionListener listener : typedSelectionListeners) {
+						listener.widgetSelected(new SelectionEvent(evt));
 					}
 				}
 			}
@@ -240,8 +243,7 @@ public class SplitButton extends Button {
 		if (listener == null) {
 			SWT.error(SWT.ERROR_NULL_ARGUMENT);
 		}
-		final TypedListener typedListener = new TypedListener(listener);
-		listeners.add(typedListener);
+		typedSelectionListeners.add(listener);
 	}
 
 	/**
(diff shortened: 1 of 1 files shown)


widgets/splitbutton/org.eclipse.nebula.widgets.splitbutton/src/org/eclipse/nebula/widgets/splitbutton/SplitButton.java
```diff
@@ -179,30 +179,29 @@ private void drawTriangle() {
...
+			} else {
+				Supplier<Event> eventFactory = () -> {
+					Event evt = new Event();
+					evt.widget = SplitButton.this;
+					evt.display = getDisplay();
+					evt.type = SWT.Selection;
+					return evt;
+				};
+				for (Listener listener : selectionListeners) {
+					listener.handleEvent(eventFactory.get());
```
laeubi: Actually it looks wrong to create a new event for each listener, @lcaron what do you think?

merks:
Unless bounds are changed that will allow this latest Nebula bundle to only installed along with latest version of SWT, this type of change is going to cause grief.

If it were "my project" I would probably try to create a utility that allowed me to work reflectively with both the old "APIs" and the new APIs so that these things continue to "just work".

widgets/splitbutton/org.eclipse.nebula.widgets.splitbutton/src/org/eclipse/nebula/widgets/splitbutton/SplitButton.java
HannesWell: It used to be like that, this is why I replicated it.
But when looking at SWT I get the impression that there is only one Event send to all listeners, unless one uses a typed listener (e.g. SelectionListener), then the one event is wrapped in a new `SelectionEvent` for each `SelectionListener`, see `TypedListener`. But the latter is probably a technical detail.
And although Events are not immutable, I think good listeners don't change events.

HannesWell:
> Unless bounds are changed that will allow this latest Nebula bundle to only installed along with latest version of SWT, this type of change is going to cause grief.

This change does not use the recent API addition in SWT (as you can see in the CI it builds with the latest SimRel) and just avoids `TypedListener` by reworking the listener handling. I assume that's the source of your concern, isn't it?
Although I don't know if this is compatible with SWT 1.0, but it shouldn't raise the requirements on the SWT version compared to the previous state.

> If it were "my project" I would probably try to create a utility that allowed me to work reflectively with both the old "APIs" and the new APIs so that these things continue to "just work".

The new APIs are only used in https://github.com/eclipse/nebula/pull/583, but there I raised the version bounds. But lets discuss there if any other strategy should be used.

merks:
Sorry, I didn't look in detail of what's being used!

widgets/splitbutton/org.eclipse.nebula.widgets.splitbutton/src/org/eclipse/nebula/widgets/splitbutton/SplitButton.java
merks: Except for "doit" when the event has that sort of thing.

HannesWell: > Except for "doit" when the event has that sort of thing.

Didn't knew that, thanks for the hint. So maybe there should be one `Event` send to all `Listeners` and a new `SelectionEvent` for each `SelectionListener`?

merks: I think that's how SWT generally does it...

laeubi: > And although Events are not immutable, I think good listeners don't change events.

Actually listeners can give some things back to the caller (e.g. as mentioned doit), there is even a field `data` mentioned for "application use" in TypedEvent, I think if one can install listeners to an SWT object one should always be "nice" and only ever modify things if the API tells so.

HannesWell:
@lcaron, @wimjongman can you please review this as well?
This only uses already existing SWT API.

widgets/splitbutton/org.eclipse.nebula.widgets.splitbutton/src/org/eclipse/nebula/widgets/splitbutton/SplitButton.java
```diff
@@ -179,30 +179,26 @@ private void drawTriangle() {
...
-				} else {
-					for (final Listener listener : listeners) {
-						final Event evt = new Event();
-						evt.widget = SplitButton.this;
-						evt.display = getDisplay();
-						evt.type = SWT.Selection;
-						listener.handleEvent(evt);
-					}
+		super.addListener(SWT.MouseDown, event -> {
+			boolean shouldShowMenu = event.x >= splitButtonAreaLeft && event.y >= splitButtonAreaTop
```
wimjongman: For readability the shouldShowMenu method should not be inlined.

wimjongman (commented):
One remark. Otherwise LGTM. I did not run it so I don't know if the widget works the same way it used to.

widgets/splitbutton/org.eclipse.nebula.widgets.splitbutton/src/org/eclipse/nebula/widgets/splitbutton/SplitButton.java
HannesWell: Ok, just reverted that part.

HannesWell:
> One remark. Otherwise LGTM. I did not run it so I don't know if the widget works the same way it used to.

I'm not familiar with Nebula in general and the `SplitButton` especially, but when I run the `SplitButtonSnippet` everything seem to work fine as far as I understood it.

HannesWell:
@wimjongman is there anything missing here or can this be submitted?

wimjongman:
Go for it. Thanks, Hannes!

HannesWell:
> Go for it. Thanks, Hannes!

I'm not a committer here, but thanks Ed for submitting.

wimjongman:
> > Go for it. Thanks, Hannes!
> 
> I'm not a committer here, but thanks Ed for submitting.

You get a special mention when we release. 🏅

PR #583, opened 2024-03-22
Replace usage of internal SWT TypedListener and SWTEventListener

HannesWell:
The `org.eclipse.swt.widgets.TypedListener` should be considered an internal class (as the doc states) and leaks the class `SWTEventListener` which also resides in an internal package.

Leverage new methods introduced in https://github.com/eclipse-platform/eclipse.platform.swt/pull/1112 in order to avoid the usage of that two classes. I don't know if there is a specific backwards compatibility policy in Nebula regarding SWT, but at the moment the new methods are only available in Eclipse I-builds. I guess therefore this change cannot be applied immediately.
In order to demonstrate it works I changed temporarily changed the target-platform of the build.

@lcaron could you please have a look at this one?

@Phillipus for your information (if you want to review this too, please do so).

### DIFF
--- a/widgets/calendarcombo/org.eclipse.nebula.widgets.calendarcombo/src/org/eclipse/nebula/widgets/calendarcombo/CustomCombo.java
+++ b/widgets/calendarcombo/org.eclipse.nebula.widgets.calendarcombo/src/org/eclipse/nebula/widgets/calendarcombo/CustomCombo.java
@@ -30,7 +30,6 @@ import org.eclipse.swt.widgets.Listener;
 import org.eclipse.swt.widgets.Menu;
 import org.eclipse.swt.widgets.Shell;
 import org.eclipse.swt.widgets.Text;
-import org.eclipse.swt.widgets.TypedListener;
 import org.eclipse.swt.widgets.Widget;
 
 /**
@@ -259,11 +258,7 @@ public class CustomCombo extends Composite {
 	 * @see #removeModifyListener
 	 */
 	public void addModifyListener(ModifyListener listener) {
-		checkWidget();
-		if (listener == null)
-			SWT.error(SWT.ERROR_NULL_ARGUMENT);
-		TypedListener typedListener = new TypedListener(listener);
-		addListener(SWT.Modify, typedListener);
+		addTypedListener(listener, SWT.Modify);
 	}
 
 	/**
@@ -294,12 +289,7 @@ public class CustomCombo extends Composite {
 	 * @see SelectionEvent
 	 */
 	public void addSelectionListener(SelectionListener listener) {
-		checkWidget();
-		if (listener == null)
-			SWT.error(SWT.ERROR_NULL_ARGUMENT);
-		TypedListener typedListener = new TypedListener(listener);
-		addListener(SWT.Selection, typedListener);
-		addListener(SWT.DefaultSelection, typedListener);
+		addTypedListener(listener, SWT.Selection, SWT.DefaultSelection);
 	}
 
 	/**
@@ -325,11 +315,7 @@ public class CustomCombo extends Composite {
 	 * @since 3.3
 	 */
 	public void addVerifyListener(VerifyListener listener) {
-		checkWidget();
-		if (listener == null)
-			SWT.error(SWT.ERROR_NULL_ARGUMENT);
-		TypedListener typedListener = new TypedListener(listener);
-		addListener(SWT.Verify, typedListener);
+		addTypedListener(listener, SWT.Verify);
 	}
 
 	void arrowEvent(Event event) {
@@ -1417,10 +1403,7 @@ public class CustomCombo extends Composite {
 	 * @see #addModifyListener
 	 */
 	public void removeModifyListener(ModifyListener listener) {
-		checkWidget();
-		if (listener == null)
-			SWT.error(SWT.ERROR_NULL_ARGUMENT);
-		removeListener(SWT.Modify, listener);
+		removeTypedListener(SWT.Modify, listener);
 	}
 
 	/**
@@ -1442,11 +1425,8 @@ public class CustomCombo extends Composite {
 	 * @see #addSelectionListener
 	 */
 	public void removeSelectionListener(SelectionListener listener) {
-		checkWidget();
-		if (listener == null)
-			SWT.error(SWT.ERROR_NULL_ARGUMENT);
-		removeListener(SWT.Selection, listener);
-		removeListener(SWT.DefaultSelection, listener);
+		removeTypedListener(SWT.Selection, listener);
+		removeTypedListener(SWT.DefaultSelection, listener);
 	}
 
 	/**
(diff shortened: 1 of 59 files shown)


laeubi:
> I guess therefore this change cannot be applied immediately.

Yes it will at least require a new SWT release

HannesWell:
> > I guess therefore this change cannot be applied immediately.
> 
> Yes it will at least require a new SWT release

Yes.

I also want to mention that, even if SWT will eventually deprecate and (re-)move the `SWTEventListener` and `TypedListener` it will not happen before at least the two year deprecation period has passed.
So it would also be possible to await the actual removal before submitting this PR (although I wouldn't mind to submit it earlier :) ) in order to avoid using tricks.

releng/org.eclipse.nebula.nebula-parent/pom.xml
```diff
@@ -33,7 +33,7 @@ Contributors:
 		<jacoco-version>0.8.9</jacoco-version>
 		<easymock-version>5.2.0</easymock-version>
 
-		<target-platform>https://download.eclipse.org/releases/latest</target-platform>
+		<target-platform>https://download.eclipse.org/eclipse/updates/4.32-I-builds/</target-platform>
```
merks: FYI, replacing the release train site with the Platform's I-builds site is going to reduce the available set of 3rd party bundles significantly, leading to resolution failures.

HannesWell: Yes I noticed that as well. But since I assumed it will take some time until this is submitted anyway and it was temporary I was lazy to look up how the repos are set up.
But now I checked it and it should work. :)

HannesWell: Now the build succeeded :)

Phillipus: After 4.32 M1 release you could change it to `https://download.eclipse.org/releases/2024-06/`?

releng/org.eclipse.nebula.nebula-parent/pom.xml
```diff
@@ -272,6 +272,11 @@ Contributors:
 			<layout>p2</layout>
 			<url>${target-platform}</url>
 		</repository>
+		<repository>
+			<id>eclipse-iBuild</id>
+			<layout>p2</layout>
+			<url>https://download.eclipse.org/eclipse/updates/4.32-I-builds</url>
+		</repository>
```
HannesWell: This should be reverted before submission.

releng/org.eclipse.nebula.nebula-parent/pom.xml
HannesWell: Yes that would be possible too, but would then later have to be changed back, if the latest is generally desired.

merks: FYI, during the end of the last release cycle SimRel produced and now maintains this:

https://download.eclipse.org/releases/milestone/

as requested here:

https://github.com/eclipse-simrel/simrel.build/issues/286

This site has the advantage that it tracks milestones and release candidates so that any problems are noticed before the release rather than after the release.

lcaron:
Thank you @HannesWell . It seems to be a big change. @wimjongman what do you think about this proposal ?

widgets/cdatetime/org.eclipse.nebula.widgets.cdatetime/src/org/eclipse/nebula/widgets/cdatetime/CDateTime.java
```diff
@@ -1527,9 +1524,7 @@ protected void postClose(Shell popup) {
 	 */
 	public void removeSelectionListener(SelectionListener listener) {
 		if (listener != null) {
-			TypedListener l = new TypedListener(listener);
-			removeListener(SWT.Selection, l);
-			removeListener(SWT.DefaultSelection, l);
+			addTypedListener(listener, SWT.Selection, SWT.DefaultSelection);
```
merks: This definitely looks wrong.

HannesWell: Oops. Looks like a C&P error.
Will fix that, thanks for spotting!

widgets/carousel/org.eclipse.nebula.widgets.carousel/src/org/eclipse/nebula/widgets/carousel/Carousel.java
```diff
@@ -165,8 +165,7 @@ public void addImage(final Image image) {
 	 * @see SelectionEvent
 	 */
 	public void addSelectionListener(final SelectionListener listener) {
-		checkWidget();
-		SelectionListenerUtil.addSelectionListener(this, listener);
```
wimjongman: @lcaron Can this be done like this?

wimjongman:
> Thank you @HannesWell . It seems to be a big change. @wimjongman what do you think about this proposal ?

When the method is officially released, we can take it.

HannesWell:
> > Thank you @HannesWell . It seems to be a big change. @wimjongman what do you think about this proposal ?
> 
> When the method is officially released, we can take it.

It was just released today. I rebased the branch and resolved all conflicts.

merks:
@wimjongman  @lcaron 

Any objections to merging this now?

widgets/calendarcombo/org.eclipse.nebula.widgets.calendarcombo/src/org/eclipse/nebula/widgets/calendarcombo/CustomCombo.java
```diff
@@ -259,11 +258,7 @@ public void add(String string, int index) {
 	 * @see #removeModifyListener
 	 */
 	public void addModifyListener(ModifyListener listener) {
-		checkWidget();
```
wimjongman, line 262: Why is the checkwidget and the null check removed?

wimjongman (changes requested):
Some queries.

widgets/calendarcombo/org.eclipse.nebula.widgets.calendarcombo/src/org/eclipse/nebula/widgets/calendarcombo/CustomCombo.java
HannesWell: `Widget.addTypedListener()` does the same already, so it would just be a duplication to do it here.

widgets/carousel/org.eclipse.nebula.widgets.carousel/src/org/eclipse/nebula/widgets/carousel/Carousel.java
HannesWell: What's the issue you see with this change? `checkWidget()` is called in `Widget.addTypedListener()`.

wimjongman:
Thanks, Hannes!

HannesWell:
Thanks for your review!

PR #585, opened 2024-04-03
bug[TW25783]: Fix Filter DateTime Column

vaibhavypatel:
Change-Id: I88e37a6ff8a89fdbaf6a8062327e983e540450b8


merks:
Could you explain what is being fixed, how it looks before your change and how it looks after?

Previously org.eclipse.nebula.widgets.xviewer.XViewerSorter.parseDatePair(String, String) was called and it used either this:
```
 public final static SimpleDateFormat format10 = new SimpleDateFormat("MM/dd/yyyy");
```
or used java.text.SimpleDateFormat.SimpleDateFormat() which is locale specific.

Now that method is no longer called and you are using "MM/dd/yyyy hh:mm a" as the only hard coded format, which seems like your own personal choice rather that something that should be hard coded in the framework...

vaibhavypatel:
@merks 
You are absolutely correct. I was not aware that other projects are making use of the XViewer. So, I fixed the code to satisfy the use-case I was trying to solve.

I resolved the issue by amending new commit. Please take a look. It is generic so that it does not require Date conversion now.
Filtering for the datetype column was not working properly. It will work after the change here gets merged. 

Thank you :)

dondunne (approved):

dondunne:
+1

PR #596, opened 2024-05-28
Disable move on double-click

laeubi:
Sometimes it could happen that if one double click on a control the slider "hangs" from then on as the control does not receive a "MousUp" event and then the follow the mouse without actually holding it down.

This now additionally registers a MouseDoubleClick handler that disable the move operation.


lcaron:
Good job as usual, thank you !

PR #603, opened 2024-06-29
Do not paint over the borders of the widget

laeubi:
Currently RectangleGroupStrategy sets the clipping to the maximum area the control will need. But in case where the widget size is larger than the area to paint (for example inside a ScrolledComposite) this will result in painting over other widgets (at least on linux).

This now do two things:

 1) save the original clipping region and recover it afterwards
 2) intersect the calculated clipping region with the original to
possibly shrink it to the are that is valid for repaint.

### DIFF
--- a/widgets/pgroup/org.eclipse.nebula.widgets.pgroup/src/org/eclipse/nebula/widgets/pgroup/RectangleGroupStrategy.java
+++ b/widgets/pgroup/org.eclipse.nebula.widgets.pgroup/src/org/eclipse/nebula/widgets/pgroup/RectangleGroupStrategy.java
@@ -35,15 +35,15 @@ import org.eclipse.swt.graphics.Region;
 public class RectangleGroupStrategy extends AbstractGroupStrategy
 {
 
-    private int vMargin = 2;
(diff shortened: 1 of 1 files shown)


wimjongman:
Did you push this without review? Please lets do the proper thing in the future.

laeubi:
@wimjongman I tried to find a suitable reviewer, but code is marked as "alpha" and no one seem to really work on this (I found some other things but I can workaround them), are you haveing any concerns in not painting over other widgets it seems totally wrong and produces some ugly artifacts at least on linux.

wimjongman:
The patch is great as usual, Christoph. Just add me or @lcaron in the future. We are mostly quick to review. The main goal is that way we are kept in the loop of changes.

laeubi:
Alright, I just don't wanted to bore you both with reviews so just let me know if it is too much and you think I should go on with something on my own :-D

wimjongman:
> Alright, I just don't wanted to bore you both with reviews so just let me know if it is too much and you think I should go on with something on my own :-D

😂sure thing.

PR #602, opened 2024-06-21
Use Widget.removeTypedListener() in CheckTableCombo

HannesWell:
Follow-up on https://github.com/eclipse/nebula/pull/583 because `CheckTableCombo` was changed since that PR was created.

@wimjongman could you please have a look?

PR #613, opened 2024-11-24
Replace usage of internal SWT TypedListener in CheckTableCombo

HannesWell:
Leverage new methods introduced in https://github.com/eclipse-platform/eclipse.platform.swt/pull/1112

I somehow missed that part in my previous work to replace references to actually internal SWT classes in
- https://github.com/eclipse/nebula/pull/583
- https://github.com/eclipse/nebula/pull/602

@lcaron can you have a look at this as well. Thanks in advance.


HannesWell:
If @lcaron is not available @wimjongman could you please have a look at this?

Once this is submitted, could you please consider to create a new release for Nebula as soon as possible, because in SWT we would like to make changes to actually internal classes that are not compatible with older versions of Nebula:
- https://github.com/eclipse-platform/eclipse.platform.swt/pull/1615#issuecomment-2496065747

wimjongman (approved):
Thanks, Hannes!

lcaron:
Hi

I will have a look.

Take care

Laurent

Le mar. 26 nov. 2024 à 18:52, Hannes Wellmann ***@***.***> a
écrit :

HannesWell:
Thank you both!

PR #615, opened 2024-11-26
Expand range of org.eclipse.rap.rwt dependency for grid.test

merks:


merks:
@lcaron 

This helps the target platform resolve the Oomph'd workspace.  I don't think anything actually builds this.

Issue #616, opened 2024-11-26
Nebula 3.2.0 releng


wimjongman:
Tracks the release engineering activities for the 3.2.0 release.

### Opening actions
- [x] Update version numbers
- [x] Set 3.2.0 Milestone release date 
- [x] Create Milestone 3.3.0 in Github (https://github.com/EclipseNebula/nebula/milestone/12)
- [x] Create Release in [Project Administration](https://projects.eclipse.org/projects/technology.birt)

### Release day - 7
- [x] Announce new release pending in discussions
- [x] Revisit all open tasks

### Release candidate (Milestone) - testing
- [x] 3.2.0 Stable Widgets
- [x] 3.2.0 Incubation Widgets

### Release day
- [x] Eclipse Release (https://download.eclipse.org/nebula/updates/)
- [x] Github Release (https://github.com/EclipseNebula/nebula/releases/tag/V3.2.0)
- [ ] Release to Maven Central

### Post Release
- [x] Create Github release (Creates Tag and Announcement in Discussions)
- [x] Update [README.MD](https://github.com/EclipseNebula/nebula/blob/master/README.md)
- [x] Update 3.2.0 Release in [Project Administration](https://projects.eclipse.org/projects/technology.nebula)
- [x] Create 3.3.0 Release in [Project Administration](https://projects.eclipse.org/projects/technology.nebula)
- [ ] Create 3.3.0 Releng Issue
- [ ] Move 3.2.0 open issues to 3.3.0

wimjongman:
Anyone knows where this comes from? 

![image](https://github.com/user-attachments/assets/f559c495-346d-4f88-98d8-8a6ef92dc96b)

merks:
From this?

https://github.com/eclipse/nebula/blob/master/.github/workflows/maven.yaml

merks:
@wimjongman 

I asked about this and it was suggested that the scanning is code because of this configuration

https://github.com/EclipseNebula/.eclipsefdn/blob/9ed684fe553bcc57f1d57d685a6f7ff9f804f067/otterdog/eclipsenebula.jsonnet#L29

The problem appears to be that the scanner is running with an older version of Java than is required by the scanner libraries.  But I'm not sure we can influence that anywhere so I think the IT staff needs to help:

https://gitlab.eclipse.org/eclipsefdn/helpdesk/-/issues/5363

- [x] Issue Fixed

merks:
@HannesWell @wimjongman 

[...]

PR #617, opened 2024-11-26
Nebula 3.2.0 releng #616

wimjongman:
Updated feature version

@merks preparing for the next 3.2.0 nebula release.


merks:
FYI, I'm a little concerned about the state of this:

https://github.com/eclipse-platform/eclipse.platform.swt/pull/1615

Mostly that seems to be problems with RAP/rwt support...

PR #624, opened 2024-12-02
fix comparison in MaskFormatter

### LINKED ISSUE #608: FormattedText.MaskFormatter Regression
Dear Nebula team,

we have run into a regression of the MaskFormatter in the FormattedText widget:

The getValue() is intended to return the actual text field value without the separator characters. But this behavior was broken in this line:

https://github.com/eclipse/nebula/blob/31f40e4e56e006f54e87eae14b23c36dbd79f09b/widgets/formattedtext/org.eclipse.nebula.widgets.formattedtext/src/org/eclipse/nebula/widgets/formattedtext/MaskFormatter.java#L153

The last comparison should also test for equality:
c == P_LHEXDIGIT

Such that the complete if condition would read:
			if (c == P_DIGIT || c == P_ALPHANUM || c == P_UPPERCASE
					|| c == P_LOWERCASE || c == P_UHEXDIGIT || c == P_LHEXDIGIT) {

Could you please fix this?

regards
Jens

claussenj:
Fix for https://github.com/EclipseNebula/nebula/issues/608


merks:
I revalidated the ECA and it seems fine now.

wimjongman (approved):
Looks good to me!

PR #625, opened 2024-12-18
Widget page migrated to Github Markdown

### LINKED ISSUE #536: Move example screenshots from https://eclipse.dev/nebula/ to this repo
For see example of the nebula widgets I have to go to https://eclipse.dev/nebula/, would it be possible to move these examples to this repo for example as EXAMPLES.md?

vogella:
Rename of images which were inconsistently named
Text adjusted to be shorter

Fixes #536


laeubi:
@vogella why not include this directly into the main readme? Just at the place where you wanted to add a link here: https://github.com/EclipseNebula/nebula/pull/626/files

vogella:
@laeubi I think it serves a better purpose as standalone document similar to the SWT snippets or swt widget page which are also not part of the readme.

@wimjongman @lcaron wdyt?

laeubi:
> similar to the SWT snippets or swt widget page which are also not part of the readme

... what does not mean we should not change it :-)

Alternatively I think it could be part of the [organization-page](https://github.com/EclipseNebula) .. my opinion is just we should show what the project offers to the user, instead of hiding this behind links one need to click.

If I look at the initial page [here](https://github.com/EclipseNebula/nebula) it looks rather technical to me and not very appealing. As nebula offers some cool widgets we should at least present some of the most prominent ones there.

wimjongman:
We have to do both. A list with all widgets and a carousel in the main page.

wimjongman:
Thanks, Lars!

wimjongman:
Oh, you added it to the main repository? Maybe not a bad idea but I also started a Github wiki. https://github.com/EclipseNebula/nebula/wiki

Maybe we should not have a Github wiki at all but just use the docs folder? What do you all think?

laeubi:
You can also use github pages (like tycho does)

vogella:
> Oh, you added it to the main repository? Maybe not a bad idea but I also started a Github wiki. https://github.com/EclipseNebula/nebula/wiki
> 
> Maybe we should not have a Github wiki at all but just use the docs folder? What do you all think?

I personally prefer the docs folder compared to wiki pages. This allows to send PR

PR #627, opened 2024-12-25
Correctly report the required size to paint the PGroup control

laeubi:
Currently if the the content of the PGroup is smaller than the title area width the text is cropped and possibly not readable because the PGroup does not report the title area with as it preferred width.

This now changes the computation so that the hints are passed to the render and then it could report the desired size.

The problem can be seen with the [PGroupSnippet1.java](https://github.com/EclipseNebula/nebula/blob/master/examples/org.eclipse.nebula.snippets/src/org/eclipse/nebula/snippets/pgroup/PGroupSnippet1.java) already here it is how it is shown before (under Linux):

pgroup_before

As one can see, even though there is enough space left, the title is abbreviated and hardly readable.

After this change it looks like this:

pgroup_after


lcaron (approved):
Nice job, congrats !

lcaron:
Thank you @laeubi for your contribution 👍

PR #629, opened 2025-01-15
Alignment for second PShelf item is broken

### LINKED ISSUE #145: PShelf in Nebula not scaling during DPI Change
During the dpi change, the texts are not correctly aligned vertically. 

In this ss: primary monitor is 100% and secondary is 150%

Image

ShahzaibIbrahim:
Alignment is broken for all n+1 items is broken because GC returns wrong height from the font metrics.

### How to Test

> Run the PShelfSnippet1 with the following VM Arguments on 200% monitor (Also happens for 100% but difference is not that much visible)

```
-Dswt.autoScale=quarter
-Dswt.autoScale.updateOnRuntime=true
```
> See if the second item is properly aligned vertically.

### Before Fix
image

### After Fix
image

### DIFF
--- a/widgets/pshelf/org.eclipse.nebula.widgets.pshelf/src/org/eclipse/nebula/widgets/pshelf/PaletteShelfRenderer.java
+++ b/widgets/pshelf/org.eclipse.nebula.widgets.pshelf/src/org/eclipse/nebula/widgets/pshelf/PaletteShelfRenderer.java
@@ -55,7 +55,6 @@ public class PaletteShelfRenderer extends AbstractRenderer {
 		return new Point(wHint,h);
 	}
 
-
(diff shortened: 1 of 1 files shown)


wimjongman (approved):
Please use blocks {}

wimjongman:
What is the status of this?

ShahzaibIbrahim:
@wimjongman the solution works but I am not sure why. That's why it's still drafted. I needed some time to figure out why it works this way.

fedejeanne:
@wimjongman  We propose to merge it as it is as it fixes the PR related issue. Would you agree on merging?

wimjongman:
I merged it. Thank you for the patch!

PR #639, opened 2025-05-19
Upgrade Tycho and use stable 2025-03 Eclipse Release in Target Platform

laeubi:
Build is currently broken due to RAP increased there version and now takes precedence over real SWT but missing some of the constants and classes.

### DIFF
--- a/.mvn/maven.config
+++ b/.mvn/maven.config
@@ -1,0 +1,1 @@
+-Dtycho-version=4.0.12
--- a/releng/org.eclipse.nebula.nebula-parent/pom.xml
+++ b/releng/org.eclipse.nebula.nebula-parent/pom.xml
@@ -20,7 +20,6 @@ Contributors:
 	<packaging>pom</packaging>
 
 	<properties>
-		<tycho-version>4.0.4</tycho-version>
 		<tycho-extras-version>${tycho-version}</tycho-extras-version>
 		<mockito-version>5.10.0</mockito-version>
 		<findbugs-version>3.0.5</findbugs-version>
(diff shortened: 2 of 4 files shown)


.mvn/maven.config
```diff
@@ -0,0 +1 @@
+-Dtycho-version=4.0.12
```
merks, line 1: It's just a comment, but I find such configuration even more difficult to find than having it in the pom.xml.  What's the advantage?

releng/org.eclipse.nebula.nebula-parent/pom.xml
```diff
@@ -33,7 +32,7 @@ Contributors:
 		<jacoco-version>0.8.9</jacoco-version>
 		<easymock-version>5.2.0</easymock-version>
 
-		<target-platform>https://download.eclipse.org/releases/milestone</target-platform>
+		<target-platform>https://download.eclipse.org/releases/2025-03</target-platform>
```
merks, line 35: This also has the disadvantage of not noticing new problems and becoming stale in just a few weeks.  What do you see as the advantage of this versus milestone or latest?

.mvn/maven.config
laeubi: Otherwise one needs to maintain the version in `pom.xml` and `extensions.xml` see https://tycho.eclipseprojects.io/doc/latest/StructuredBuild.html

releng/org.eclipse.nebula.nebula-parent/pom.xml
laeubi: > What do you see as the advantage of this versus milestone or latest?

We have a working build. Also Nebula is actually not meant to be using any latest features, so there is no need to update this. As it is a system property one can of course have a build that set it to milestones/latest/.... in an additional build if we want to have such checks.

lcaron:
Thank you Christoph !

PR #640, opened 2025-05-19
Use minimal latest release repositories for builds

merks:
- The primary goal is to ensure that we don't unintentionally pick up RAP/RWT dependencies from the SimRel repository.
- Make the corresponding changes to the URLs used for the Oomph setup.
- Update the jar signer version to 1.5.2.
- Don't use versions for org.w3c.dom package imports so they can resolve to the JDK.
- Remove bogus logging dependencies.
- Avoid all bundle imports referencing RWT or RAP; use package imports instead.

### DIFF
--- a/oomph.setup
+++ b/oomph.setup
@@ -175,7 +175,7 @@
         <repository
             url="https://download.eclipse.org/tools/gef/classic/releases/latest"/>
         <repository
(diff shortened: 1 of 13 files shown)


laeubi (approved):
Sounds good!

PR #641, opened 2025-05-31
Create Colors without a display

laeubi:
Since a while now swt allows creating colors without a display. Nebula still uses the old way and attaching dispose listeners that are run needlessly (since colors are no longer need dispose).

This replace all calls to colors with a display by their plain variant without.


wimjongman (approved):
Thank you, Christoph!

PR #642, opened 2025-05-31
[rounded toolbar] Add support for setting the border color

laeubi:
Currently one can not control the color of the border drawn around toolbar items.

This now adds a way to set the border color for a toolbar to any color.

Screenshot from the snippet:
grafik


wimjongman (approved):
Thank you! 🙏🙏

PR #643, opened 2025-06-07
[Grid] Redraw column when text is updated

laeubi:
Currently setting the text of a column is only effective if some kind of redraw is triggered elsewhere.

This now checks if the text has changed and redraw the bounds of the column to update it immediately.


lcaron:
Thank you :)

PR #644, opened 2025-06-08
[rounded toolbar] Allow setting the active color of the toolbar

laeubi:
Currently it is only possible to change the default color of the toolbar using a rather unusual constructor argument. SWT widget are generally only have one constructor with the parent and the style.

This now deprecates the constructor that was used to init the default color and replace it with a proper getter. Additionally it is now also possible to change the color of the active state (when a button is pressed).

This is how it look like with the adjusted snippet:

grafik


wimjongman (approved):
🙏❤️

lcaron:
Good job, thank you Christoph !

PR #645, opened 2025-06-30
[grid] Add basic fixed column support

laeubi:
Currently Nebula Grid does not offer a feature know as "fixed columns" where one can set some columns to be displayed regardless of the scrolling.

This now adds a very basic first support for such columns, even though it does not work for all grid features it allows to fix a column in a simple grid. 

@wimjongman there are likely more things to do but I'd like to keep it small for the moment as the Grid class itself is a > 10k LOC code monster already, so there will come more PRs for other features. To be honest most likely such feature I use my self, but of course everyone is invited to help improving this :-)

This does:

1. Add API to mark column fixed
2. Support painting of fixed column headers and table content

It currently does not support for example (but might not be exhaustive) that should come next:

1. Change size of fixed columns when scrolling is done
2. Edit fixed columns in a fixed way
3. ...


wimjongman (approved):
Thank you, Christoph!

PR #646, opened 2025-07-01
Fix Chips is using wrong SWT style constant

laeubi:
Currently the Chips uses `SWT.PUSH` in case where actually `SWT.TOGGLE` behavior is provided.

This now fixes the problem by replacing `SWT.PUSH` with `SWT.TOGGLE` and adjust the snippet accordingly.

@wimjongman @lcaron Even though it is kind of "breaking" change, I think we should still do that, the current usage of `SWT.PUSH` is confusing compared to what buttons and toolbars do here. Beside that, if the PR is accepted I plan to implement real `SWT.PUSH` behavior for chips, so users will need to just replace `SWT.PUSH` > `SWT.TOGGLE` if they want really toggle behavior, so I incremented the minor version only. If we think its better to increment the major version let me know.


widgets/chips/org.eclipse.nebula.widgets.chips/src/org/eclipse/nebula/widgets/chips/Chips.java
```diff
@@ -40,7 +40,7 @@
  * <dt><b>Styles:</b></dt>
  * <dd>SWT.CLOSE</dd>
  * <dd>SWT.CHECK</dd>
- * <dd>SWT.PUSH</dd>
+ * <dd>SWT.TOGGLE</dd>
```
wimjongman, line 43: What if current consumers use SWT.PUSH? Their application will break after upgrading to this version, right? I'm fine with replacing TOGGLE for PUSH but the old API should still be respected. This means that you have to also catch PUSH and assume they mean TOGGLE.

wimjongman (changes requested):
Thanks, Christoph. I have one remark. Please take a look.

wimjongman:
Sorry, I did not read your description. No, we cannot just replace PUSH with TOGGLE. In this version you can replace PUSH but still catch the style and translate it internally to TOGGLE.

We have to increment the major version when you implement the real PUSH.

laeubi:
@wimjongman I now made it that way that `SWT.TOGGLE` and `SWT.PUSH` do the same thing.

wimjongman (approved):
Brill! Thank you for your flexibility.

wimjongman:
You seem to use our widgets frequently. Very nice. What is the application you are making?

laeubi:
> You seem to use our widgets frequently. Very nice. What is the application you are making?

I'm maintaining some [custom RCP application](https://läubisoft.gmbh/en/) (you can [find one here](https://labratry.de/en/) for example) and often the default SWT widget are not very well suited, so if there are any demands I usually look first at Nebula if there is something to use. Next thing is that often one want to customize things what also often is not possible with the default widgets.

So as I see any issues or missing features I try to improve / add things here so others hopefully can benefit from that as well and even if its is sometimes tricky its easier than write something completely new and even better than showing poor UI to the user ;-)

PR #647, opened 2025-07-06
Support PUSH chips style

laeubi:
Currently one can not have a real push chips style (only poorly emulate this by setting all colors to the same).

This now adds a true PUSH state to Chips so one even can see when pushing the chip because it changes its color.


wimjongman (approved):
Nice and clean. Thank you.

PR #648, opened 2025-07-06
Make a disabled SwitchButton appear visually disabled

laeubi:
Currently the only effect when disabling a switch button is that the user can't change the state anymore, but there is no way to visually distinguish a disabled from an enabled button. This makes the UI feel buggy from the users point of view as the button does not work for an unknown reason.

This now makes the button using the default disabled colors when it is disabled appearing grey like other controls when disabled.

## Before

grafik

## After

grafik


wimjongman (approved):
Very nice! Thanks.

PR #650, opened 2025-07-16
[grid] Add support for filling columns

laeubi:
Currently if the view port of the grid is larger than the sum of its columns width we currently show a gray empty space (or white in the content). This does not only look like wasted space but often there actually is some content that would deserve more space if there is any.

This now adds support for a new style on GridColumn (`SWT.FILL`) that marks the column as one that wants to consume extra available space. These column now gets stretched if there is space left, if more than one such column exists the space is distributed equally between them.


laeubi:
@wimjongman I plan to move this if no one objects, just need to rebase/resolve conflict now ...

PR #653, opened 2025-07-17
Avoid using DPIUtil for computeSizePixels

merks:
https://github.com/EclipseNebula/nebula/issues/651


merks:
@lcaron 

As I described in https://github.com/EclipseNebula/nebula/issues/651#issue-3238557026 I think it's misguided to compute the size in pixels because getBorderWidth returns a size in points according to its Javadoc and its implementation.

laeubi (approved):
Yes it sounds resonable to not require any DPI util here

laeubi:
Build is green so lets try it out!

PR #654, opened 2025-07-19
[grid] Fix cell text is invisible on selected lines with fixed colums

laeubi:
Currently when one has a fixed column and selects a line then scrolls the cell text becomes invisible. This is because we effectively paint twice and the empty cell renderer thinks we have no more columns and overwrite in some cases.

This now passes a flag to the call to skip the empty painting as we have done this anyways with the previous call.


laeubi:
As this is actually a bug from a previous PR and no one objected I merge this now.

wimjongman:
Thanks, Christoph.

PR #663, opened 2025-08-29
Avoid copying the GridItem.children array repeatedly

Tarmean:
The GridItem.getItems() method performs a defensive copy. Because newItem() repeatedly invoked getItems(), this creates a quadratic runtime when adding many children to the same node.

Solution: Use GridItem.getItemCount() and GridItem.getItem() where possible.

Most other uses of GridItem.getItems() could be removed by having a GridItem.iterItems() method, which is not covered here

### DIFF
--- a/widgets/grid/org.eclipse.nebula.widgets.grid/src/org/eclipse/nebula/widgets/grid/Grid.java
+++ b/widgets/grid/org.eclipse.nebula.widgets.grid/src/org/eclipse/nebula/widgets/grid/Grid.java
@@ -7957,16 +7957,16 @@ public class Grid extends Canvas {
 				index = rootItems.get(index).getRowIndex();
 			}
 		} else if (!root) {
-			if (index >= item.getParentItem().getItems().length || index == -1) {
+			if (index >= item.getParentItem().getItemCount() || index == -1) {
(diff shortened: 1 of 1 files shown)


laeubi (approved):
Yes this is a problem with the code that it often tries to avoid local variables and repeatedly call the same methods over and over again.

Using getItemcount looks fine for me, one could also thing about making the array package private access to it can be accessed more efficiently.

widgets/grid/org.eclipse.nebula.widgets.grid/src/org/eclipse/nebula/widgets/grid/Grid.java
```diff
@@ -7957,16 +7957,16 @@ int newItem(final GridItem item, int index, final boolean root) {
 				index = rootItems.get(index).getRowIndex();
 			}
 		} else if (!root) {
-			if (index >= item.getParentItem().getItems().length || index == -1) {
+			if (index >= item.getParentItem().getItemCount() || index == -1) {
```
laeubi:
```suggestion
			if (index == -1 || index >= item.getParentItem().getItemCount()) {
```

wimjongman (approved):
Thanks @Tarmean!

Issue #662, opened 2025-08-26
RichTextEditor get SWTException on Eclipse Platform 2025-03 & 2025-09


azoitl:
While some problems with the new platform are fixed e.g., #631 we still experience an issue using the RichtTextEditor from the latest milestone (https://download.eclipse.org/nebula/updates/milestone/latest). I'm not sure if this needs to be fixed here or on SWT?

<details>
<summary>Unhandled event loop exception</summary>
### Stack Trace:
```
org.eclipse.swt.SWTException: getAllOptions is not defined
	at org.eclipse.swt.browser.Edge.evaluateInternal(Edge.java:960)
	at org.eclipse.swt.browser.Edge.evaluate(Edge.java:933)
	at org.eclipse.swt.browser.WebBrowser.evaluate(WebBrowser.java:405)
	at org.eclipse.swt.browser.Browser.evaluate(Browser.java:731)
	at org.eclipse.swt.browser.Browser.evaluate(Browser.java:680)
	at org.eclipse.nebula.widgets.richtext.RichTextEditor$3.completed(RichTextEditor.java:285)
	at org.eclipse.swt.browser.Edge.lambda$43(Edge.java:1176)
	at org.eclipse.swt.widgets.RunnableLock.run(RunnableLock.java:40)
	at org.eclipse.swt.widgets.Synchronizer.runAsyncMessages(Synchronizer.java:132)
	at org.eclipse.swt.widgets.Display.runAsyncMessages(Display.java:4133)
	at org.eclipse.swt.widgets.Display.readAndDispatch(Display.java:3749)
	at org.eclipse.e4.ui.internal.workbench.swt.PartRenderingEngine$5.run(PartRenderingEngine.java:1151)
	at org.eclipse.core.databinding.observable.Realm.runWithDefault(Realm.java:339)
	at org.eclipse.e4.ui.internal.workbench.swt.PartRenderingEngine.run(PartRenderingEngine.java:1042)
	at org.eclipse.e4.ui.internal.workbench.E4Workbench.createAndRunUI(E4Workbench.java:153)
	at org.eclipse.ui.internal.Workbench.lambda$3(Workbench.java:678)
	at org.eclipse.core.databinding.observable.Realm.runWithDefault(Realm.java:339)
	at org.eclipse.ui.internal.Workbench.createAndRunWorkbench(Workbench.java:583)
	at org.eclipse.ui.PlatformUI.createAndRunWorkbench(PlatformUI.java:173)
[...]

PR #664, opened 2025-08-29
Workaround for Race Condition on Creating Browserfunctions for Edge #662

azoitl:
On the edge browser engine it can happen that the browser functions registered by the rich text editor are not yet available when the document is parsed. By setting the template url after the browser functions are registered and sleeping the thread a bit the race condition went away.

Addresses: https://github.com/EclipseNebula/nebula/issues/662 See also:
https://github.com/eclipse-platform/eclipse.platform.swt/issues/2449

### DIFF
--- a/widgets/richtext/org.eclipse.nebula.widgets.richtext/src/org/eclipse/nebula/widgets/richtext/RichTextEditor.java
+++ b/widgets/richtext/org.eclipse.nebula.widgets.richtext/src/org/eclipse/nebula/widgets/richtext/RichTextEditor.java
@@ -248,40 +248,11 @@ public class RichTextEditor extends Composite {
 
 		browser.setUrl(templateURL.toString());
 
-		browserFunctions.add(new ModifyFunction(browser, "textModified"));
-		browserFunctions.add(new KeyPressedFunction(browser, "keyPressed"));
-		browserFunctions.add(new KeyReleasedFunction(browser, "keyReleased"));
-		browserFunctions.add(new FocusInFunction(browser, "focusIn"));
-		browserFunctions.add(new FocusOutFunction(browser, "focusOut"));
-		browserFunctions.add(new JavaExecutionStartedFunction(browser, "javaExecutionStarted"));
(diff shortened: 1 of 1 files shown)


laeubi:
>  By setting the template url after the browser functions are registered and sleeping the thread a bit the race condition went away.

This looks like maybe waiting for a domready event is missing somewhere...

azoitl:
> > By setting the template url after the browser functions are registered and sleeping the thread a bit the race condition went away.
> 
> This looks like maybe waiting for a domready event is missing somewhere...

How I understood the discussion at https://github.com/eclipse-platform/eclipse.platform.swt/issues/2449 and https://github.com/eclipse-platform/eclipse.platform.swt/issues/20#issuecomment-1089252533 it is something bigger. But I'm not see deep into SWT code that I can say anything about it.

laeubi:
@azoitl pleas [see here](https://github.com/eclipse-platform/eclipse.platform.swt/discussions/1274#discussioncomment-9975050)

> I haven't had time to reproduce with a snippet... but, in case it helps anyone in the meantime, let me just note that this idea I mentioned of registering a ProgressListener and then constructing the BrowserFunction in the method org.eclipse.swt.browser.ProgressListener.completed(ProgressEvent) seems to consistently eliminate the problem.

The point is that unless the Browser is not loaded the Dom might not be ready to access, so it seems much cleaner to do it that way than hoping for a random timeout to work for everyone. So maybe something similar can work here as well?

azoitl:
@laeubi thx for the link and the explanation. Took me a bit to understand. But I tried that and I can report that the original problem is gone on windows and introduced no additional problems on Linux. So this may even be the better fix independent of any future SWT changes.

laeubi (approved):
This looks much better now, that's a usual pitfall with Javascript that some operations need careful waiting for async operations being done.

Sadly the current Browser API is all not prepared for that causing multiple issues.

azoitl:
As this is less a workaround then a nice fix now. Any chances to get this merged? Anything I need to fix/change?

laeubi:
If no one object and the CI is already green I'll merge this then soon.

azoitl:
Thx a lot!

PR #669, opened 2025-10-07
Fix memory leak and font disposal issues in FloatingText widget

vogella:
The FloatingText widget's doSetLabelText() method could trigger "widget is disposed" exceptions. Multiple calls would create new Font objects without disposing the previous ones, and the code incorrectly disposed system-managed fonts via
fLabel.getFont().dispose().

### DIFF
--- a/widgets/floatingtext/org.eclipse.nebula.widgets.floatingtext/pom.xml
+++ b/widgets/floatingtext/org.eclipse.nebula.widgets.floatingtext/pom.xml
@@ -18,7 +18,7 @@ Contributors:
 	<parent>
 		<groupId>org.eclipse.nebula</groupId>
 		<artifactId>floatingtext</artifactId>
-		<version>1.0.0-SNAPSHOT</version>
+		<version>1.0.1-SNAPSHOT</version>
 	</parent>
 
 	<artifactId>org.eclipse.nebula.widgets.floatingtext</artifactId>
--- a/widgets/floatingtext/org.eclipse.nebula.widgets.floatingtext.feature/feature.xml
+++ b/widgets/floatingtext/org.eclipse.nebula.widgets.floatingtext.feature/feature.xml
@@ -2,7 +2,7 @@
 <feature
       id="org.eclipse.nebula.widgets.floatingtext.feature"
       label="Nebula Floating Text Widget"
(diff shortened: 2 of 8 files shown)


widgets/floatingtext/org.eclipse.nebula.widgets.floatingtext/pom.xml
```diff
@@ -18,7 +18,7 @@ Contributors:
 	<parent>
 		<groupId>org.eclipse.nebula</groupId>
 		<artifactId>floatingtext</artifactId>
-		<version>1.0.0-SNAPSHOT</version>
+		<version>1.0.1-SNAPSHOT</version>
```
merks, line 21: The version in the MANIFEST.MF needs to match this version.

merks (changes requested):

merks:
I'm confused though, is this not change the same file as in?

https://github.com/EclipseNebula/nebula/pull/670/

vogella:
> I'm confused though, is this not change the same file as in?
> 
> #670

#670 enabled me to use this widget under Linux, this is suppossed to fix the non-disposed error, I may have mixed the PR with the different commits, I check tonight and update PR accordingly.

vogella:
Sorry, took me much longer to react

widgets/floatingtext/org.eclipse.nebula.widgets.floatingtext/pom.xml
merks: Something got "smart" and removed the version from the artifact because it's the same as the parent, but now the parent version is not changed while the artifact needs a version element with 1.0.1.

image

So you need to change the parent version back to what it was and add a version element for the attribute.

vogella: That "something" was me :-o , the AI did it correctly, but I though that a parent update was not necessary. Do I also need the feature and snippet version update?

vogella: Looks like it, AI was right, I was wrong. :-(

merks (approved):
Looks fine.

PR #670, opened 2025-10-07
Floating widget gc leak

vogella:


merks:
Oh yes, the early return would cause a leak....

PR #673, opened 2025-10-31
TableCombo widget does not close the popup shell

### LINKED ISSUE #672: TableCombo widget does not close the popup shell after selecting an item from the popup shell
Our project upgraded from version 3.1.1 to 3.2.0 and found a issue in the TableCombo widget.  When you open the drop down for the TableCombo widget and select an item, the popup shell does not disappear.

I'm able to reproduce the issue using the Nebula's Examples product, so it's not our usage of the widget in our project.

I traced it back to commit 907f49d1a0efeb892b0ff6a172b2aba5397daa3c where the problem was introduced.

It's reproducible on Windows 11 and Linux. Here is a screen shot of the example app with the issue.

Image

lyncis-elanor:
fixes issue #672

### DIFF
--- a/widgets/tablecombo/org.eclipse.nebula.widgets.tablecombo/src/org/eclipse/nebula/widgets/tablecombo/TableCombo.java
+++ b/widgets/tablecombo/org.eclipse.nebula.widgets.tablecombo/src/org/eclipse/nebula/widgets/tablecombo/TableCombo.java
@@ -925,7 +925,8 @@ public class TableCombo extends Composite {
 		table.setParent(tempShell);
 
 		shellsToBeDisposed.add(tempShell);
-
+		
+		popup.setVisible(false);
 		popup = tempShell;
 	}
 
--- a/widgets/tablecombo/org.eclipse.nebula.widgets.tablecombo.css.feature/feature.xml
+++ b/widgets/tablecombo/org.eclipse.nebula.widgets.tablecombo.css.feature/feature.xml
@@ -2,7 +2,7 @@
 <feature
       id="org.eclipse.nebula.widgets.tablecombo.css.feature"
       label="Nebula Table Combo CSS"
-      version="1.2.0.qualifier"
+      version="1.3.0.qualifier"
       provider-name="Eclipse Nebula">
 
    <description url="https://www.eclipse.org/nebula">
@@ -14,7 +14,7 @@
    </license>
 
    <requires>
-      <import feature="org.eclipse.nebula.widgets.tablecombo.feature" version="1.2.0.qualifier"/>
+      <import feature="org.eclipse.nebula.widgets.tablecombo.feature" version="1.3.0.qualifier"/>
    </requires>
 
    <plugin
--- a/widgets/tablecombo/org.eclipse.nebula.widgets.tablecombo.css.feature/pom.xml
+++ b/widgets/tablecombo/org.eclipse.nebula.widgets.tablecombo.css.feature/pom.xml
@@ -20,7 +20,7 @@ Contributors:
 	<parent>
 		<groupId>org.eclipse.nebula</groupId>
 		<artifactId>tablecombo</artifactId>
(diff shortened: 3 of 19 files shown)


merks:
@ShahzaibIbrahim

This does look like a correct fix to the regression to me.

N1k145:
@merks I noticed that the SimRel (2025-09) and the current Nebula Release both contain a Table Combo 1.2.0, but in SimRel there is a build from 2024 that does not contain the issue, where as in the Nebula Release a version from 2025 is contained with the issue. Maybe the version number needs to be fixed

merks:
@N1k145 

I expect there will always be a qualifier increase no matter what.  

@lcaron  @wimjongman 

I don't now the general approach for how versions are managed in nebula.

---

Looking at SimRel, I see it's present there because it's used by Oomph and contributed via Oomph's repository:

image

Oomph is getting it from here:

image

Probably we (Oomph) should be using 

https://download.eclipse.org/nebula/updates/release/latest

But given the widget is broken, that seems like a bad plan at the moment.

merks:
Unless I hear from someone by the end of the day, I will merge this PR.

Thanks @lyncis-elanor for contributing!!

N1k145:
@merks Im also expecting a qualifier increase, but im also expecting atleast a minor version increase

merks:
@N1k145 

Indeed, in general it would be nice if (and it is expected that) during each release cycle that bundles that are changed will update something more than just the qualifier.  But, without some tool to check this and enforce this it's not so likely to happen properly.  I see nothing configured to do baseline checking, baseline replacement, nor an API baseline configured and checked; API tools actually don't help with maintaining micro versions anyway.   Oomph has tooling for managing micro-version increments, which I used in EMF, Oomph, and DataTools, but I hate to force others to use stuff that's not really so well documented...

So probably we just need to rely on people remembering to do the right things at the right time...

merks:
For example, in DataTools if I change anything that affects the binary contents there will be an error message with a quick fix to increment the version, configured to always increment the minor version in this case because no API tools is in place to help with that:

image

The released versions are recorded and after a release I delete that file which then creates new digests of the project contents.

image

In this ways I can manage the complexity with minimal effort.  (But I have way too much to do already.)

N1k145:
@merks I can totaly understand that, and I think its also not your responsibility to fix that for every project. 
Maybe that is something the maintainer of Nebula can have an eye on, that would be very helpfull for the users downstream, like us. When we analysed the issue we first wonderd where it comes from as it was the same version number as in our previous version, where the issue was not present.

wimjongman:
Totally right. @lyncis-elanor are you able to bump the minor version number of the tablecombo to 1.3.0?

Update all occurrences of 1.2.0 to 1.3.0 in this directory and subdirectories
* https://github.com/EclipseNebula/nebula/tree/master/widgets/tablecombo

wimjongman:
@copilot can you give the version bump a shot

merks:
I'll just merge this and then do a very increment.

widgets/tablecombo/org.eclipse.nebula.widgets.tablecombo/src/org/eclipse/nebula/widgets/tablecombo/TableCombo.java
```diff
@@ -925,7 +925,8 @@ private void recreatePopup() {
 		table.setParent(tempShell);
 
 		shellsToBeDisposed.add(tempShell);
-
+		
+		popup.setVisible(false);
```
merks, line 929: Or does someone want to deal with the copilot thing?  It seems to me the loop before also assumes the popup isn't null.

lyncis-elanor:
@wimjongman @merks I bumped the version everywhere I found it.

wimjongman (approved):

wimjongman:
Thank you, @lyncis-elanor!

PR #675, opened 2025-12-16
Fix GridColumn resize for FILL columns - detect and resize next column

laeubi:
Currently if one resizes a fill column nothing seem to happen. This is because the fill column increases its size but as it is filled this has no visual effect (mostly).

This now changes the behavior by resizing instead the column on the right of the fill column so that actual sizing occurs.


lcaron (approved):
Perfect job (as usual). Thank you for the fix

PR #687, opened 2026-02-13
Fixes #686 - [RichText] Add support for monitor specific scaling

### LINKED ISSUE #686: [RichText] Add support for monitor specific scaling
With the ongoing efforts to support monitor specific scaling and the planned deprecation of `Display#getDPI()`, the `RichTextEditor` and the `RichTextViewer` need to be updated.

There is a [discussion in the NatTable repository](https://github.com/eclipse-nattable/nattable/discussions/164) and also a discussion in a [ticket in SWT](https://github.com/eclipse-platform/eclipse.platform.swt/issues/2974)

It turned out that the inconsistencies are likely issues related to "magic numbers" in the richtext code and some issues with the ckeditor on Edge. At least it looks like this and can be solved there.

fipro78:
This PR adds support for monitor specific scaling in the richtext widget.

It uses the new `Shell#getZoom()` API from the upcoming 2026-03 release, so it should not be merged until the release is done.

There is also one open topic related to conversion from a pixel value to point for the font creation.

### DIFF
--- a/widgets/richtext/org.eclipse.nebula.widgets.richtext.example/META-INF/MANIFEST.MF
+++ b/widgets/richtext/org.eclipse.nebula.widgets.richtext.example/META-INF/MANIFEST.MF
@@ -4,7 +4,7 @@ Bundle-Name: Nebula Rich Text Editor Examples
 Bundle-SymbolicName: org.eclipse.nebula.widgets.richtext.example
 Bundle-Version: 1.0.0.qualifier
 Bundle-Vendor: Eclipse Nebula
-Bundle-RequiredExecutionEnvironment: JavaSE-11
+Bundle-RequiredExecutionEnvironment: JavaSE-21
 Require-Bundle: org.eclipse.jface;bundle-version="3.10.1",
  org.eclipse.nebula.widgets.richtext;bundle-version="1.0.0"
 Export-Package: org.eclipse.nebula.widgets.richtext.example
--- a/widgets/richtext/org.eclipse.nebula.widgets.richtext.example/META-INF/MANIFEST.MF
+++ b/widgets/richtext/org.eclipse.nebula.widgets.richtext.example/META-INF/MANIFEST.MF
@@ -2,7 +2,7 @@ Manifest-Version: 1.0
 Bundle-ManifestVersion: 2
 Bundle-Name: Nebula Rich Text Editor Examples
 Bundle-SymbolicName: org.eclipse.nebula.widgets.richtext.example
-Bundle-Version: 1.0.0.qualifier
+Bundle-Version: 1.7.0.qualifier
 Bundle-Vendor: Eclipse Nebula
 Bundle-RequiredExecutionEnvironment: JavaSE-21
 Require-Bundle: org.eclipse.jface;bundle-version="3.10.1",
--- a/widgets/richtext/org.eclipse.nebula.widgets.richtext/src/org/eclipse/nebula/widgets/richtext/RichTextEditor.java
+++ b/widgets/richtext/org.eclipse.nebula.widgets.richtext/src/org/eclipse/nebula/widgets/richtext/RichTextEditor.java
@@ -1,5 +1,5 @@
 /*****************************************************************************
- * Copyright (c) 2015, 2025 CEA LIST.
+ * Copyright (c) 2015, 2026 CEA LIST.
  *
  *
  * This program and the accompanying materials
@@ -87,10 +87,10 @@ public class RichTextEditor extends Composite {
 
 	private final List<BrowserFunction> browserFunctions = new ArrayList<>();
 
-	private final ListenerList modifyListener = new ListenerList(ListenerList.IDENTITY);
-	private final ListenerList keyListener = new ListenerList(ListenerList.IDENTITY);
-	private final ListenerList focusListener = new ListenerList(ListenerList.IDENTITY);
-	private final ListenerList javaCallbackListener = new ListenerList(ListenerList.IDENTITY);
+	private final ListenerList<ModifyListener> modifyListener = new ListenerList<>(ListenerList.IDENTITY);
+	private final ListenerList<KeyListener> keyListener = new ListenerList<>(ListenerList.IDENTITY);
+	private final ListenerList<FocusListener> focusListener = new ListenerList<>(ListenerList.IDENTITY);
+	private final ListenerList<JavaCallbackListener> javaCallbackListener = new ListenerList<>(ListenerList.IDENTITY);
 
 	private final RichTextEditorConfiguration editorConfig;
 
@@ -98,6 +98,8 @@ public class RichTextEditor extends Composite {
 	private Point mouseDragPosition;
 
 	private boolean handleFocusChanges = true;
+	
+	private int[] containerEdgeDiffs = null;
 
 	/**
 	 * Key of the system property to specify a fixed directory to unpack the ckeditor resources to.
@@ -219,6 +221,15 @@ public class RichTextEditor extends Composite {
 		if (embedded) {
 			embeddedShell = new Shell(parent.getShell(), SWT.MODELESS);
 			embeddedShell.setLayout(new FillLayout());
+			
+			// The Edge implementation is traversing the ESC key from the browser to the shell,
+			// which leads to closing the shell. We need to avoid this to be able to reuse the
+			// RichTextEditor control in embedded mode.
+			embeddedShell.addTraverseListener(e -> {
+				if (e.keyCode == SWT.ESC) {
+					e.doit = false;
+				}
+			});
 		}
 
 		// remove styles that are not relevant for the browser
@@ -272,13 +283,30 @@ public class RichTextEditor extends Composite {
 					browserFunctions.add(new BrowserFunction(browser, "updateDimensions") {
 						@Override
 						public Object function(final Object[] arguments) {
-							// width and height +2 because the editor is 1 pixel
+							
+							int argWidth = ((Double) arguments[0]).intValue();
+							int argHeight = ((Double) arguments[1]).intValue();
+							
+							// width and height + containerEdgeDiffs because the editor is
 							// smaller than the browser container on every side
+							int newWidth = argWidth + (containerEdgeDiffs != null ? containerEdgeDiffs[0] : 2);
+							int newHeight = argHeight + (containerEdgeDiffs != null ? containerEdgeDiffs[1] : 2);
+							
+							String updateOnRuntime = System.getProperty("swt.autoScale.updateOnRuntime", "false"); //$NON-NLS-1$ //$NON-NLS-2$
+					        if (!Boolean.parseBoolean(updateOnRuntime)) {
+					        	argWidth = ScalingHelper.getZoomedValue(argWidth, embeddedShell != null ? embeddedShell.getZoom() : 100);
+					        	argHeight = ScalingHelper.getZoomedValue(argHeight, embeddedShell != null ? embeddedShell.getZoom() : 100);
+
+					        	newWidth = ScalingHelper.getZoomedValue(newWidth, embeddedShell != null ? embeddedShell.getZoom() : 100);
+					        	newHeight = ScalingHelper.getZoomedValue(newHeight, embeddedShell != null ? embeddedShell.getZoom() : 100);
+					        }
+
+					        Rectangle currentBounds = getBounds();
 							setInlineContainerBounds(
-									getBounds().x,
-									getBounds().y,
-									((Double) arguments[0]).intValue() + 2,
-									((Double) arguments[1]).intValue() + 2);
+									currentBounds.x,
+									currentBounds.y,
+									newWidth,
+									newHeight);
 
 							// also repaint the parent control to avoid
 							// rendering glitches while resizing
@@ -287,6 +315,38 @@ public class RichTextEditor extends Composite {
 								getParent().update();
 							}
 
+							return super.function(arguments);
+						}
+					});
+
+					browserFunctions.add(new BrowserFunction(browser, "initialDimensions") {
+						@Override
+						public Object function(final Object[] arguments) {
+							if (embeddedShell != null && containerEdgeDiffs == null) {
+								int initialWidth = ((Double)arguments[0]).intValue();
+								int intialHeight = ((Double)arguments[1]).intValue();
+
+								Rectangle embeddedShellBounds = embeddedShell.getBounds();
+
+								String updateOnRuntime = System.getProperty("swt.autoScale.updateOnRuntime", "false"); //$NON-NLS-1$ //$NON-NLS-2$
+						        if (!Boolean.parseBoolean(updateOnRuntime)) {
+						        	initialWidth = ScalingHelper.getZoomedValue(initialWidth, embeddedShell.getZoom());
+						        	intialHeight = ScalingHelper.getZoomedValue(intialHeight, embeddedShell.getZoom());
+
+						        	int widthDiff = embeddedShellBounds.width - initialWidth;
+						        	int heightDiff = embeddedShellBounds.height - intialHeight;
+
+									containerEdgeDiffs = new int[] { 
+											ScalingHelper.getUnzoomedValue(widthDiff, embeddedShell.getZoom()), 
+											ScalingHelper.getUnzoomedValue(heightDiff, embeddedShell.getZoom()) };
+						        } else {
+						        	int widthDiff = embeddedShellBounds.width - initialWidth;
+						        	int heightDiff = embeddedShellBounds.height - intialHeight;
+						        	
+					        		containerEdgeDiffs = new int[] { widthDiff, heightDiff };
+						        }
+							}
+							
 							return super.function(arguments);
 						}
 					});
@@ -788,16 +848,14 @@ public class RichTextEditor extends Composite {
 	 *            the new height for the receiver
 	 */
 	void setInlineContainerBounds(final int x, final int y, int width, int height) {
-		width = ScalingHelper.convertHorizontalPixelToDpi(width);
-		height = ScalingHelper.convertVerticalPixelToDpi(height);
 		resizedBounds = new Rectangle(x, y, width, height);
 		if (embeddedShell != null) {
 			final Point shellLocation = embeddedShell.getLocation();
 			embeddedShell.setBounds(
 					shellLocation.x,
 					shellLocation.y,
-					width + 2,
-					height + 2);
+					width,
+					height);
 		}
 		else {
 			super.setBounds(x, y, width, height);
@@ -814,7 +872,7 @@ public class RichTextEditor extends Composite {
 	 *         editor resize minimum in case the editor was created with {@link SWT#MIN}
 	 */
 	protected int getMinimumHeight() {
-		return ScalingHelper.convertVerticalPixelToDpi(200);
+		return ScalingHelper.getZoomedValue(200, embeddedShell != null ? embeddedShell.getZoom() : 100);
 	}
 
 	/**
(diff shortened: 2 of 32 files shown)


widgets/richtext/org.eclipse.nebula.widgets.richtext.example/META-INF/MANIFEST.MF
```diff
@@ -4,7 +4,7 @@ Bundle-Name: Nebula Rich Text Editor Examples
 Bundle-SymbolicName: org.eclipse.nebula.widgets.richtext.example
 Bundle-Version: 1.0.0.qualifier
```
merks: You should have at least some some minimal version change here.

merks (commented):
I think this:

https://github.com/EclipseNebula/nebula/blob/328cd098b4d89c673a228202f6e407f2af6ac1a6/.github/workflows/maven.yaml#L12

and this

https://github.com/EclipseNebula/nebula/blob/328cd098b4d89c673a228202f6e407f2af6ac1a6/Jenkinsfile#L14

need attention with your Java 21 upgrade.

laeubi:
> There is also one open topic related to conversion from a pixel value to point for the font creation.

Last time I looked into that topic was: Don't do it... actually at Linux I can configure a custom scaling factor for fonts making anything that tries to "adjust" fonts simply fail.

Next SWT/OS lies at you (well ... it do best effort)... so you get a logical DPI not the real hardware DPI... and so on.

So the best one can do is having adapt to system font sizes and maybe add a preference for the user to define an own scaling factor.

If you need Fonts matching an exact bounds, the best is to compute it dynamically using the Font Metrics and then possibly adjust it incrementally.

widgets/richtext/org.eclipse.nebula.widgets.richtext.example/META-INF/MANIFEST.MF
fipro78: I suppose you mean the version of the richtext bundle, not the examples. Did that.

fipro78:
> I think this:
> 
> https://github.com/EclipseNebula/nebula/blob/328cd098b4d89c673a228202f6e407f2af6ac1a6/.github/workflows/maven.yaml#L12
> 
> and this
> 
> https://github.com/EclipseNebula/nebula/blob/328cd098b4d89c673a228202f6e407f2af6ac1a6/Jenkinsfile#L14
> 
> need attention with your Java 21 upgrade.

Thanks for the hint. As the upcoming Eclipse release requires Java 21, this needs to be adapted here also.

@laeubi 
I only understand your statement partially. It doesn't really make sense to me in the context of the richtext use case. Probably because I did not clarify the use case. 

With CKEditor you can edit and format text in HTML. For the text formatting, the corresponding style elements are added to the tags. And I need to convert these style informations to SWT. The font height is set in px, but Fonts in SWT need to be created with point values. So there is no existing font that would match here at that time from which I could retrieve FontMetrics. At least I would not be aware of something like that. If you have an idea how to solve this in a better way, please let me know.

widgets/richtext/org.eclipse.nebula.widgets.richtext.example/META-INF/MANIFEST.MF
merks: If you change **anything** in any bundle you should increment that bundle's version:

image

laeubi:
> I only understand your statement partially. It doesn't really make sense to me in the context of the richtext use case. 

Then maybe you can give an example of where exactly this conversion is needed.

> With CKEditor you can edit and format text in HTML

So far so good :-)

> The font height is set in px

Where? Html supports a wide range of font size definitions including px, pt, em, ...

> Fonts in SWT need to be created with point values. So there is no existing font that would match here at that time from which I could retrieve FontMetrics

If you can create Fonts you have a Display and you can create a GC and get FontMetrics from that....

widgets/richtext/org.eclipse.nebula.widgets.richtext.example/META-INF/MANIFEST.MF
fipro78: Nobody cares about the examples. They are not published and only used for development testing AFAIK. That is why nobody ever mentioned them in 11 years.

But ... you are completely right, and just because nobody cared so far doesn't mean it is correct. ;)
I updated the version. :)

merks: I didn't actually look at the context around this.   😁

widgets/richtext/org.eclipse.nebula.widgets.richtext/src/org/eclipse/nebula/widgets/richtext/RichTextEditor.java
```diff
@@ -291,6 +319,38 @@ public Object function(final Object[] arguments) {
...
+						@Override
+						public Object function(final Object[] arguments) {
+							if (embeddedShell != null && containerEdgeDiffs == null) {
+								int initialWidth = ((Double)arguments[0]).intValue();
+								int intialHeight = ((Double)arguments[1]).intValue();
+
+								Rectangle embeddedShellBounds = embeddedShell.getBounds();
+
+								String updateOnRuntime = System.getProperty("swt.autoScale.updateOnRuntime", "false"); //$NON-NLS-1$ //$NON-NLS-2$
+						        if (!Boolean.parseBoolean(updateOnRuntime)) {
```
HeikoKlare: I have to admit that I do not understand why the distinction between monitor-specific scaling disabled/enabled has to be made here. I would expect that if you need to do something special for non-100% in monitor-specific scaling, you will probably need to do the same with monitor-specific scaling disabled at zooms of 175% and higher (but that's just a guess because I do not know why these different calculations are necessary here).

Anyway, in case the distinction is necessary, I propose to use `getDisplay.isRescalingAtRuntime()` instead of evaluating the system property. It should give you an even "more correct" value (as you can change the system property after the display with its shells has been created).

fipro78:
@laeubi 

The HTML is generated by CKEditor, and I don't have much influence on how it is generated. For example if a text is selected to have a size of 20, the generated HTML looks like this

```
<p><span style="font-size:20px">Eclipse</span></p>
```

In the `RichTextPainter` basically the HTML is parsed to generate instructions for painting. In one place the font styling is parsed, and if the font size is provided as px, it needs to be converted to points.

https://github.com/EclipseNebula/nebula/blob/328cd098b4d89c673a228202f6e407f2af6ac1a6/widgets/richtext/org.eclipse.nebula.widgets.richtext/src/org/eclipse/nebula/widgets/richtext/RichTextPainter.java#L631

I use instructions to process the tags accordingly, and the font style is applied via `span` in ckeditor. So the font according to the styling is created in the `SpanStylePaintInstruction`

https://github.com/EclipseNebula/nebula/blob/328cd098b4d89c673a228202f6e407f2af6ac1a6/widgets/richtext/org.eclipse.nebula.widgets.richtext/src/org/eclipse/nebula/widgets/richtext/painter/instructions/SpanStylePaintInstruction.java#L81

You can use the [RichTextViewerExample](https://github.com/EclipseNebula/nebula/blob/master/widgets/richtext/org.eclipse.nebula.widgets.richtext.example/src/org/eclipse/nebula/widgets/richtext/example/RichTextViewerExample.java) to see it in action. With the current implementation, the selected font size in the browser reflects correctly in the viewer. At least it does on Windows with different scalings (or at least it did).

@HeikoKlare 
The distinction is necessary because the Javascript returns different values than SWT. Probably because the Javascript execution does not know about the autoscaling setting. So if updateOnRuntime is false on a 150% scaling monitor, I get the 100% values from the javascript callback, but the 150% values via `Shell.getBounds()`. If I set updateOnRuntime to true, I get the 150% values from the javascript callback and from `Shell.getBounds()`. Why? I have no idea! Took me quite a while to find and handle this. I would expect that the pixel values should be the same in the browser and the shell, as the browser renders inside the shell, and therefore they have the same dimensions. But actually they don't.

Thanks for the hint about `getDisplay.isRescalingAtRuntime()`, I will adapt this and do some further testing.

@merks 
Is there anything else I have to modify for the Java 21 update? Seems the build still fails because there is no Java 21 available.

HeikoKlare:
> If I set updateOnRuntime to true, I get the 150% values from the javascript callback and from `Shell.getBounds()`. Why? I have no idea! Took me quite a while to find and handle this. I would expect that the pixel values should be the same in the browser and the shell, as the browser renders inside the shell, and therefore they have the same dimensions. But actually they don't

I a quite sure this is not caused by updateOnRuntime but by the SWT autoscale mode that is (implicitly) changed when enabling updateOnRuntime. Precisely, without updateOnRuntime the autoscale mode `integer` is used whereas with updateOnRuntime enabled it defaults to `quarter`. You can see this by starting the application with `-Dswt.autoScale.updateOnRuntime=false -Dswt.autoScale=quarter` where you should make the same observations regarding the sizes as with updateOnRuntime enabled.
So all this is caused by the special autoscale handling of SWT that was introduced like 10 years ago, which used to only know about 100% and 200% for bounds, sizes etc. but exact scales for the fonts, which led to weird results before monitor-specific scaling was available.
And it means that with updateOnRuntime set to false, for 175% and above you should get scaled values from the Javascript as well and have to deal with them (which does not seem to be covered by the current implementation). 

Related to this, the size of the shell does not look as intended to me when having updateOnRuntime set to false.

This is the size at 150%:
image

And this at 175%:
image

So the shell seems to be autoscaled to 100%/200%.

merks:
> @merks
> Is there anything else I have to modify for the Java 21 update? Seems the build still fails because there is no Java 21 available.

Sorry, I forgot about your question.  Change these both to JavaSE-21

image

fipro78:
@HeikoKlare 
> So the shell seems to be autoscaled to 100%/200%.

That is an interesting observation. I tried to find the reason, but actually it seems to be an issue of the `Browser` somehow. The shell is resized so it is able to show the whole content. This is done in the following way:

1. The embedded shell is created with the minimum width and height. The numbers are defined so that at least one content row and the toolbar can be shown. The `Browser` is created with a `FillLayout` to fill the whole available space.
2. Once the CKEditor instance is ready, the editor is maximized to fill the whole space. I now call the `initialDimensions` browser function to be able to calculate the diff between ckeditor and shell bounds.

With updateOnRuntime set to true, everything works as intended. Setting it to false, the values reported by the Javascript only provide 100% or 200% values, not the ones expected by the dimensions of the `Shell` or the `Browser`.

The minimum width is 370, the minimum height is 200.

With display scaling set to 175% I get:

updateOnRuntime=false
```
embeddedShellBounds=Rectangle {392, 141, 648, 350}, initialWidth=724, intialHeight=398
```

updateOnRuntime=true
```
embeddedShellBounds=Rectangle {500, 281, 648, 350}, initialWidth=631, intialHeight=349
```

With display scaling set to 150% I get:

updateOnRuntime=false
```
embeddedShellBounds=Rectangle {848, 659, 555, 300}, initialWidth=353, intialHeight=197
```

updateOnRuntime=true
```
embeddedShellBounds=Rectangle {375, 194, 555, 300}, initialWidth=537, intialHeight=298
```

The bounds of the `Browser` control match the bounds of the embedded shell. Why the dimensions reported by the browser engine do not match with updateOnRuntime=false is totally unclear to me. It is especially strange as with updateOnRuntime=true the values are correct, and the browser engine itself should not be aware of the SWT autoscaling settings, or do I misunderstand that? Any ideas? Could that be an issue in the `Browser` implementation?

I also tried to use the `window.innerHeight` and `window.innerWidth` values, which show the same values, of course with a slight difference because of the bounds.

HeikoKlare:
> Why the dimensions reported by the browser engine do not match with updateOnRuntime=false is totally unclear to me. It is especially strange as with updateOnRuntime=true the values are correct, and the browser engine itself should not be aware of the SWT autoscaling settings, or do I misunderstand that? Any ideas? Could that be an issue in the `Browser` implementation?

To me it looks as if the dimensions reported by the browser might be correct, just the shell is initialized/updated with an unintended size on SWT side due to wrong calculation based on the shell zoom.
Some clarification: two shells with the same pixel dimensions on a 150% screen will have different logical dimension in SWT in case one uses updateOnRuntime=true and the other uses updateOnRuntime=false. This is _not_ because of the monitor-specific scaling enabled with updateOnRuntime itself but because of the fact that a different autoscale mode ("integer" instead of "quarter" is used).
With updateOnRuntime (and autoscale mode "quarter") activated, dimension values will appear more "expected" because there is a linear conversion between pixel-based size and logical size in SWT based on the actual zoom. With autoscale mode "integer" (the one we used to have for years as default), zoom values are rounded to multiples of 100, which is why a shell with pixel size (200, 200) will have a logical size of (200, 200) on 150% monitor as the autoscaling zoom of SWT is 100 in that case, and a logical size (100, 100) on a 175% monitor as the autoscaling zoom of SWT is 200.

In my understanding, the values coming from the browser via JavaScript should probably always be in pixels. The values used by SWT (in particular for setting the shell size/bounds and retrieving those values) are in logical SWT points. They need to be transformed into a common representation consistently, which does not always seem to be the case. For example, the `getMinimumWidth()` method is used when calculating the bounds of the shell, but `getMinimumWidth()` seems to calculate a pixel-based value whereas the shell expects a logical point-based value. Maybe such issues also exist in other calculation leading to the unintended results.

That's the reason why I think the calculations based on `Shell.getZoom()` currently added in this PR may not be correct as well (and might be the reason for the issue we discuss here). `Shell.getZoom()` always returns the "native" zoom of the shell, i.e., 150 in case of a 150% monitor (see also the Javadoc of `Shell.getZoom()`). When using the "old" autoscaling mode (updateOnRuntime=false and thus autoscaling mode "quarter"), SWT will calculate everything with a zoom of 100.
As an example, the browser may have a pixel size of (150, 150) on a 150% monitor. Then the current calculation (based on `shell.getZoom()` will apply the native zoom of 150 and thus give you a logical size of (100, 100). SWT however treats this as pixel value (100, 100) when using autoscaling mode "integer" as it uses an autoscaling zoom of 100. In my understanding, the pixel values coming from the browser need to be converted with the SWT autoscaling zoom and not the native zoom to make everything consistent.

In addition, I am not sure if the conversion in `ScalingHelper.getZoomValue()` is conceptually correct. Currently, the values passed from the browser via JavaScript (assumed to be pixel-based) are scaled _up_ (multiplied) with the zoom, but to convert into logical SWT coordinates you would have to scale _down_ (divide) by the zoom. So the calculation is inverted to what you would expect.

As a side note on the complexity of these autoscaling calculations: there were good reasons for doing this "rounding" of zooms to multiples of 100 (10 years ago or so when HiDPI support was first introduced), but it also introduces quite some complexity on dealing with logical coordinates. That's why I already mentioned at other places that I would be in favor of getting rid of thoe autoscaling capabilities as soon as possible. But currently, we have to deal with that situation.

**Proposal:** What I would try to do to clean this up is to convert every value involved in the calculations to a SWT logical, point-based value and then only operate on such point-based values. Maybe that's even the intention right now, but in my understanding the calculation don't properly reflect that (as explained above). In addition, I am still quite sure that any distinction based on "updateOnRuntime" is not correct. Differences are caused by the used autoscaling mode and are independent from "updateOnRuntime" (you should see the same differences when comparing `-Dswt.autoScale=quarter -Dswt.autoScale.updateOnRuntime=false` with `-Dswt.autoScale=integer -Dswt.autoScale.updateOnRuntime=false`).
In case there is no other way of properly applying SWT autoscaling for the conversion of pixel-based values coming from the browser, you might keep the distinction based on "updateOnRuntime" but then need to do the following:
- "updateOnRuntime==true": do the calculations based on `shell.getZoom()`
- "updateOnRuntime==false": do the calculations based on `DPIUtil.getDeviceZoom()`

This would be reasonable because with updateOnRuntime we enforce an autoscaling zoom that equals the native zoom whereas without updateOnRuntime you usually have the "integer" scaling mode and the zoom value rounded to the next multiple of 100 is provided via `DPIUtil.getDeviceZoom()` (and since there is only a single system-wide zoom, using that static value is okay). But note that `DPIUtil.getDeviceZoom()` is internal and must not be used in case of `updateOnRuntime==true`.

fipro78:
Thanks a lot for the detailed explanation. But to be honest, I am more confused than before. You said I have to use `Shell.getZoom()` but it seems to be not correct here. You say I can use `DPIUtil.getDeviceZoom()`, but that method is not API.

And actually I don't really understand what you mean exactly with

> pixel values coming from the browser need to be converted with the SWT autoscaling zoom and not the native zoom

and 

> convert every value involved in the calculations to a SWT logical, point-based value and then only operate on such point-based values

What does that mean? How can I do that? Where do I get this magic factor from? 

I really have no idea what I should do now to correct this, to become a good Eclipse/SWT citizen again.

HeikoKlare:
> What does that mean? How can I do that? Where do I get this magic factor from?

To be honest, I would have been surprised if you were able to easily derive a proper solution from my explanations. I don't know either how to make this work properly (i.e., support all kinds of autoscaling modes in a fully API-compliant way). The existing SWT autoscaling concept is nice when you just consume SWT or implement custom stuff that integrates well with SWT (such as just rendering with a GC). When it comes to integrating other applications (such as browsers or in case of our product LibreOffice) it can become painful when you have to synchronize positions/sizes of both, because those external applications don't know about SWT's autoscaling, so you have to break the scaling encapsulation of SWT. That means you either need to access internals (such as `DPIUtil`) or you may rely on implicit contracts (such as that with monitor-specific scaling enabled you can assume a linear relation between logical point values and pixel values depending on a shell's zoom i.e., `shell.getZoom()`). That's what I referred to with:
> pixel values coming from the browser need to be converted with the SWT autoscaling zoom and not the native zoom

and

> convert every value involved in the calculations to a SWT logical, point-based value and then only operate on such point-based values

That said, I am a bit surprised that the current `RichTextEditor` implementation worked at all for fractional zooms on Windows. Taking a look at the existing implementation of `RichTextEditor` and the usage of `convertHorizontalPixelToDPI`, that method used `display.getDPI()` which included the "correction" for SWT autoscaling and somehow produced the expected conversion result. But I don't get how this is properly related to the sizes reported from the browser via JavaScript (and I would expect them to be in pixels but then just scaling with a DPI-based value would be incorrect). Still it somehow seems to have worked as expected (or if it did not, no one noticed it).

Maybe, if the calculation based on `display.getDPI()` worked as expected, it could still produce the expected results now (even though I do not understand why, but as said, I do not understand why it ever has worked).
Or maybe you just keep the existing implementation for whatever scenario it worked (such as monitor-specific scaling disabled, maybe only autoscaling mode "integer") until you finally drop support for that configuration and have a toggle to a new implementation based on `shell.getZoom()` whenever appropriate (i.e., in my understanding, when the current autoscaling mode is `quarter` or `exact` but maybe its sufficient to stick to monitor-specific being enabled as that information is available via public API).

fipro78:
@HeikoKlare 

> To be honest, I would have been surprised if you were able to easily derive a proper solution from my explanations.

Thanks a lot for that statement. I was already doubting myself.

> That said, I am a bit surprised that the current RichTextEditor implementation worked at all for fractional zooms on Windows.

To be honest, I haven't tested all the fractional zooms so far. But what I can tell is, that `getDPI()` returns the "point-modified" value, so I get a 96 for example on 150% with autoscale=quarter. Not sure if that is the reason why it worked so far.

I finally understood that it is not the updateOnRuntime flag, but rather the autoscale configuration. And that the default value is different depending on the updateOnRuntime flag. I think I have found a solution. It at least worked in my different tests. Not sure if it is now correct, but at least I could not force the issue anymore. Only for the font calculation from pixel to point it seems I still need the `getDPI()` method. Although there seems to be a method available in the platform. But of course not public API.

widgets/richtext/org.eclipse.nebula.widgets.richtext.example/META-INF/MANIFEST.MF
fipro78: Updated bundle and feature versions and increased the BREE

fipro78:
@merks and @HeikoKlare 

I have updated the PR a while ago and triggered the build just now to build against the latest Eclipse Platform. From my end I would say it looks good now. Not sure if there are some issues hidden for specific configurations, but at least it doesn't break for me with these changes.

wimjongman (approved):
Looks good to me. Let's take it well before the release so that we have time to find issues. 

Thanks to all involved, this took a big chunk of your time and it is greatly appreciated. 💕

PR #697, opened 2026-04-28
Add TriStateSwitchButton with two active states and off

laeubi:
I recently have had the demand for a button that can have tri-state behavior. This is what I came up with based on the existing `SwitchButton` and the copilot agent:

grafik


laeubi:
@wimjongman @lcaron any concerns? Seems one can't move forward anymore without at least one review.

wimjongman (approved):
Apologies for the delay. Thanks for the patch.

wimjongman:
I added documentation here: https://github.com/EclipseNebula/nebula/wiki/Nebula-Switch-Buttons

lcaron:
Great work, thank you Christoph for your contribution !

PR #698, opened 2026-06-14
Expand range for jakarta.annotation to include 3.x

merks:


merks:
@wimjongman  @fipro78 @laeubi 

The build are failing like this:

> [ERROR]   Software being installed: org.eclipse.nebula.widgets.richtext.example.e4 1.7.0.qualifier
> [ERROR]   Missing requirement: org.eclipse.nebula.widgets.richtext.example.e4 1.7.0.qualifier requires 'java.package; jakarta.annotation [2.1.0,3.0.0)' but it could not be found: See log for details

So this is needed.

wimjongman (approved):

laeubi:
> @wimjongman @fipro78 @laeubi
> 
> The build are failing like this:
> 
> > [ERROR]   Software being installed: org.eclipse.nebula.widgets.richtext.example.e4 1.7.0.qualifier
> > [ERROR]   Missing requirement: org.eclipse.nebula.widgets.richtext.example.e4 1.7.0.qualifier requires 'java.package; jakarta.annotation [2.1.0,3.0.0)' but it could not be found: See log for details
> 
> So this is needed.

I just wanted to note that this is due to the removal of 2.0 from the eclipse updatesite. So generally one would have been adding the missing dependency as well, but extending the range seems sufficient as well.

merks:
> I just wanted to note that this is due to the removal of 2.0 from the eclipse updatesite. So generally one would have been adding the missing dependency as well, but extending the range seems sufficient as well.


Extending the range is generally what the Platform did (which is why 2.x is now missing from the Platform's update site) so Nebula is in good company:

image

